PR #91, opened 2019-07-08
Make the cmake nicer looking

viktorjo:
Removed util as a built library and made the linking procedure simpler. Also removed the supervisorcontrol module from the building procedure, since it is not commonly used.

Also changed the name of the program to Maestro and not TEServer :)


LukasWikander (changes requested):
The runServer.sh script (in repo root dir) should also be updated with the new name of the TEServer executable.

LukasWikander (changes requested):

konglobemeralt (changes requested):
Personally I Think that the math library should be named something more descriptive than "m".  
EDIT: When I looked at this the first time I thought m was our meastroMath and not math.h so guess there is nothing we can do

konglobemeralt (approved):

PR #151, opened 2019-11-21
Enabled ScenarioControl in the runServer script in preparation for the bug hunting session

LukasWikander:
We should test all modules at the same time to catch as much as possible


konglobemeralt (approved):
:+1:

PR #180, opened 2019-12-02
ObjectControl not crashing if an object suddenly disconnect from server.

sepast:
Not trying to do close(*socket) on the socket that has disconnected.


LukasWikander (changes requested):
Discussed this in person. uiRecvMonitor causes the crash. Also, Jenkins complains :cry:

LukasWikander (approved):
Looks good! Two small changes, plus whatever Jenkins is complaining about (I'm guessing that you can just run formatSourceFiles.sh)

PR #194, opened 2020-01-09
Feature fix straggling problems

LukasWikander:
In addition to the commit comments, I also made ObjectControlSendTRAJ force the socket to be temporarily blocking since problems occurred when sending really large trajectories.


konglobemeralt (approved):
Looks good. Can add that this was tested in the office  

RIP moni  :disappointed:

konglobemeralt (approved):

PR #195, opened 2020-01-09
Added 15 minute timeout to entire Jenkins pipeline

LukasWikander:
Does what the name describes


konglobemeralt (approved):
> Does what the name describes

Big if true

PR #214, opened 2020-01-27
Move T&A and INSUP to iso22133

LukasWikander:
I've also moved a bunch of typedefs into the .c file from the .h file


viktorjo (commented):
I have some small comments right now, which would be nice to have discussed before approval. I will continue with testing the server now.

PR #215, opened 2020-01-31
Feature move asp handler functions to util

LukasWikander:


viktorjo (commented):
I feel like the title of this pull-request is a bit missleading. You are moving the ASP handler functions to iso22133 files right? I am just wondering if there is something I am missing? In the insertions and deletions it seems like nothing had been added to the util.h and util.c files.

PR #276, opened 2020-07-17
Refactor data dictionary and remove GSD from it

LukasWikander:
We have roughly 22 getXasU32, 22 getXasC8, 22 setXasU32, 22 setXasC8, 22 initX etc. which all do very similar things and the same code is repeated 22 times for each function.

This pull request replaces all of the functions which used to get a variable in GSD coupled to test.conf with a single function template. Same for set, and for the singleton variables stored in shmem.

I am still working on figuring out a non-repetitive code pattern for the variables stored in shmem with more than one element.


sepast (commented):
I nice draft, let's try to use this one.

PR #287, opened 2020-08-20
Disable packet timeout function by setting MaxPacketsLost in test.con…

sepast:
…f to 0.

When testing, it's annoying when Maestro disconnects to objects due to MONR packet loss. This is in this feature possible to disable by setting MaxPacketsLost to 0 in test.conf,


LukasWikander (changes requested):
Setting max missing packets to zero is the preferred way of having it according to ISO (i.e. we should be receiving MONR at least at 100 Hz frequency with no missing packets). In other words, setting it to zero should be the most strict setting, not the least strict.

In case you find this annoying I would suggest making it possible to set the setting to UINT32_MAX and set it really high to "disable" it.

PR #316, opened 2021-01-07
Journal control includes files from all days

LukasWikander:
The main functionality is in `insertNonBookmarked()`, which inserts whole files (first - last line) in the correct time span which are not already tracked due to the start/stop bookmarks.

Took the opportunity to also clean up the code some by making a class containing the whole collection of tracked journals.


alfaro01 (commented):
Added some comments after skimming. Is it necessary to test it and is there a simpler way than running it for multiple days?

oscjoha (approved):

PR #337, opened 2021-03-30
Improved the manual python testing scripts to enable testing of Maest…

viktorjo:
…ro without using the GUC


konglobemeralt (commented):
We should probably add a ticket in the future to integrate the testing with the new ROS-server instead of keeping this.

PR #371, opened 2021-08-27
replaced FileLengthI32 with strlen(functionReturnName)

konglobemeralt:
A small bug to fix for the DownloadTrajs command


sepast (commented):
Please check in feature_TrajDirDownload for the solution we agreed about.

sepast (commented):
Check the comment

PR #373, opened 2021-09-07
Check if BTS is allowed 

konglobemeralt:
I think the code is pretty self explanatory. Currently the BTS-module however does not use the transmitter IDs for anything and as such this is a standalone method. My thinking however is that this method will be run two times during BTS

First to verify that the objects have run their originalTrajectory (Can they run the BTS-traj?) and secondly to verify that they are in a position to run their original traj,


LukasWikander (approved):
Renaming of a function + calling a method which does the boilerplate for you.

LukasWikander (approved):

alfaro01 (approved):

PR #377, opened 2021-09-09
Test for reverse traj

LukasWikander:
Simple test reversing a traj twice, expecting to get the same result back. Could you help me out here @konglobemeralt ?


sepast (approved):

SebastianSvanland:
@LukasWikander Is this PR still relevant?

LukasWikander:
Yes it's still relevant, but we need to fix the bug before we merge this.

PR #406, opened 2021-12-16
Fixed usage of visualization server gsd dd functions

LukasWikander:


sepast (approved):

konglobemeralt (approved):
I did a quick scan of this since it appears to already be approved but not merged. Is this done @sepast ? I can merge it then so we avoid conflict with the GSD stuff

PR #409, opened 2021-12-20
Remove gsd from external supervisor settings.

sepast:


konglobemeralt (approved):
This looks good. Is this the last one?

PR #413, opened 2022-01-14
Improvements to manual testing of the ISO-protocol

viktorjo:


LukasWikander (approved):

LukasWikander:
No longer relevant. In case this is needed in the future a suggestion is to make an application in the ISO 22133 repo, so we don't need to maintain separate implementations.

PR #417, opened 2022-02-03
Feature systemcontrol ros

victorjarlow:
Some progress on converting relativekinematics and systemcontrol from mqueue to ros publisher/subscriber


LukasWikander (commented):
I will continue the review on systemcontrol.cpp at a later time. For now, here are the other comments.

LukasWikander (approved):

PR #450, opened 2022-05-09
Feature per object sub pub

victorjarlow:
Each TestObject now publishes its own monr messages on the channel /maestro/object_n/object_monitor. Each object also receives (chunks of) trajs on the topic /maestro/object_n/trajectory, where n is the ID of the object. 

When ObjectControl creates a TestObject, it also initializes a monrPub and a trajSub for the object, which it provides as arguments to the constructor of the TestObject. The pub and sub are shared pointers and get destroyed whenever the TestObject gets destroyed.


konglobemeralt (approved):
I think it looks good so I will put down an approve, but maybe you can show me some of the details of how to actually use it tomorrow?

PR #465, opened 2022-10-11
Feature - Sending trajectory data in MONR-message

samuelthoren:
Changes for this pull request:

- Loading configurations file when initialising.
- Saving trajectory points.
- Calculating three points from the trajectories: 
1. A point on the trajectory nearest to the object.
2. A point in the near distance.
3. A point in the far distance. These points can then be used for the driver model.
- Sending the data as an OSI-message.
- Fixed and added some documentation.

### DIFF
--- a/launch/maestro_launch.py
+++ b/launch/maestro_launch.py
@@ -56,12 +56,12 @@ def generate_launch_description():
             executable='journal_control',
             name='journal_control'
         ),
-        Node(
-            package='rviz2',
-            namespace='maestro',
-            executable='rviz2',
-            name='rviz2'
-        ),
+        # Node(
+        #     package='rviz2',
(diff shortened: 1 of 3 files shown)


victorjarlow (changes requested):
As we discussed offline, it is more fitting to move the specific far-near point calculation to the DriverModel (DM) and let this Maestro OSI-adapter just handle the task of sending chunks of trajectories to a reciever (which might not even be a driver model infact).

LukasWikander:
@samuelthoren should we close this PR? Seems you've made a lot of updates later on that override these changes.

samuelthoren:
> @samuelthoren should we close this PR? Seems you've made a lot of updates later on that override these changes.

Yes, we can close this

PR #508, opened 2023-02-17
Trajectorylet streamer 2.0

LukasWikander:
It is updated with esmini setup and can publish the while trajectory if configured with parameter chunk_duration 0.0
image


samuelthoren (changes requested):
I can't build, got the following errors:
image

PR #515, opened 2023-02-22
Updated ML scenario generation script to have 6 objects

LukasWikander:


victorjarlow:
Due to how the control tower is situated on Multilane, and with what we learned about the line-of-sight issues with humanetics, it is probably better if mondeo and humanetics swap lanes.

PR #522, opened 2023-02-28
added a minimum time for which a scenario is run when parsing traject…

victorjarlow:
…ories


LukasWikander (approved):
Added some more stuff @victorjarlow

LukasWikander (approved):

PR #524, opened 2023-03-02
Feature foxglove publishing

victorjarlow:
To try the feature in this PR:
1) sudo apt install ros-foxy-foxglove-msgs ros-foxy-sensor-msgs
2) Start atos
3) Go to https://studio.foxglove.dev
4) Open a new connection using Rosbridge to ws://localhost:9090
5) Import the layout in plugins/foxglove
6) Import the extension in plugins/foxglove

extension (message converter) sourcecode is available here: https://github.com/RI-SE/monr_to_foxyglove_pose/tree/main/src

image


common/trajectory.cpp
```diff
@@ -10,6 +10,7 @@
 #include <algorithm>
 #include "regexpatterns.hpp"
 #include "trajectory.hpp"
+#include "util.h" // xyz2llh
```
LukasWikander, line 13: Could you take this opportunity to move xyz2llh into a separate hpp file? Adding util to trajectory will add a lot of unwanted includes, also to other files #including this one.

common/trajectory.cpp
```diff
@@ -90,6 +91,34 @@ nav_msgs::msg::Path Trajectory::toPath() const
 	return path;
 }
 
+foxglove_msgs::msg::GeoJSON Trajectory::toGeoJSON(double llh_0[3]) const {
```
LukasWikander, line 94: We are writing C++ code here :) std::vector saves you the trouble of passing and checking the size of the array as well as the pointer validity, which you would otherwise need to do.
```suggestion
foxglove_msgs::msg::GeoJSON Trajectory::toGeoJSON(const std::vector<double>& llh_0) const {
```

modules/EsminiAdapter/src/esminiadapter.cpp
```diff
@@ -143,7 +143,6 @@ void EsminiAdapter::onStaticInitMessage(
 		RCLCPP_ERROR(me->get_logger(), e.what());
 		return;
 	}
-	me->InitializeEsmini();
```
LukasWikander, line 146: I would like to avoid moving this back here. Having it in the callback opens for race conditions when object control requests esmini data on init.

plugins/foxglove/Map and 3D layout.json
```diff
@@ -0,0 +1,94 @@
...
+        "far": 5000
+      },
+      "followMode": "follow-pose",
+      "scene": {},
+      "transforms": {},
+      "topics": {
+        "/atos/object_1/object_monitor": {
+          "visible": true,
+          "type": "arrow"
+        },
```
LukasWikander, line 36: Only monitor for one of the objects?

plugins/foxglove/Map and 3D layout.json
```diff
@@ -0,0 +1,94 @@
...
+        "/atos/object_4/path": {
+          "visible": true
+        },
+        "/atos/object_5/path": {
+          "visible": true
+        },
+        "/atos/object_6/path": {
+          "visible": true
+        }
```
LukasWikander, line 54: Would be nice to add the same colors that Samuel added for rviz paths and position arrows, to be able to separate the objects.

plugins/foxglove/Map and 3D layout.json
```diff
@@ -0,0 +1,94 @@
...
+        "poseEstimateYDeviation": 0.5,
+        "poseEstimateThetaDeviation": 0.26179939
+      }
+    },
+    "map!2s6yzns": {
+      "center": {
+        "lat": 57.777722684128044,
+        "lon": 12.781342864036562
+      },
+      "customTileUrl": "",
```
LukasWikander, line 72: We have a tile server you can use here, containing the high res photo. Ask @oscjoha

LukasWikander (changes requested):
The initializeEsmini call back into the callback creates a race condition we need to avoid.

LukasWikander:
Also, we should make a node converting object monitor to the necessary standard display types to make it possible to view foxglove in a tablet or on a phone.

victorjarlow:
> Also, we should make a node converting object monitor to the necessary standard display types to make it possible to view foxglove in a tablet or on a phone.

In addition to this, we can host our own version of https://studio.foxglove.dev/ publically, then we can pre-load the client with any extension we want, without limitation.

Another benefit is that we are not dependent on the uptime of the foxglove webserver.

LukasWikander:
> > Also, we should make a node converting object monitor to the necessary standard display types to make it possible to view foxglove in a tablet or on a phone.
> 
> In addition to this, we can host our own version of https://studio.foxglove.dev/ publically, then we can pre-load the client with any extension we want, without limitation.
> 
> Another benefit is that we are not dependent on the uptime of the foxglove webserver.

As long as we will not end up in the situation where we need to build packages specific for the devices we want to connect. The big development time bonus switching from Qt / GUC to this would be lost.

common/CMakeLists.txt
```diff
@@ -93,6 +94,9 @@ ament_target_dependencies(${ATOS_COMMON_TARGET}
 set_target_properties(${ATOS_UTIL_TARGET} PROPERTIES
 	PUBLIC_HEADER ${CMAKE_CURRENT_SOURCE_DIR}/util.h
 )
+set_target_properties(${ATOS_UTIL_TARGET} PROPERTIES
+	PUBLIC_HEADER ${CMAKE_CURRENT_SOURCE_DIR}/util/coordinateutils.hpp
+)
```
LukasWikander, line 99:
```suggestion
set_property(TARGET ${ATOS_UTIL_TARGET} APPEND PROPERTY
	PUBLIC_HEADER ${CMAKE_CURRENT_SOURCE_DIR}/util/coordinateutils.hpp
)
```

modules/EsminiAdapter/inc/esminiadapter.hpp
```diff
@@ -39,6 +38,7 @@ class EsminiAdapter : public Module {
 	ROSChannels::Init::Sub initSub;
 	ROSChannels::Start::Sub startSub;
 	std::unordered_map<uint32_t,ROSChannels::Path::Pub> pathPublishers;
+	std::unordered_map<uint32_t,ROSChannels::GeoJSON::Pub> geoJSONPublishers;
```
LukasWikander, line 41:
```suggestion
	std::unordered_map<uint32_t,ROSChannels::GeoJSON::Pub> gnssPathPublishers;
```
?

modules/EsminiAdapter/src/esminiadapter.cpp
```diff
@@ -20,6 +20,8 @@
 #include "rclcpp/wait_for_message.hpp"
 #include "trajectory.hpp"
 #include "string_utility.hpp"
+#include "util.h"
```
LukasWikander, line 23: Possible to
```suggestion
```
?

modules/ObjectControl/src/objectlistener.cpp
```diff
@@ -65,7 +68,15 @@ void ObjectListener::listen() {
...
 				JournalRecordMonitorData(&objData);
-				obj->publishMonr(createROSMessage(monr));
+
+				// Publish a ROS monr message
+				auto rosMonr = createROSMessage(monr);
+				obj->publishMonr(rosMonr);
+
+				// Publish a NavSatFix message
+				auto origin = obj->getOrigin();
+				double llh_0[3] = {origin.latitude_deg, origin.longitude_deg, origin.altitude_m};
```
LukasWikander, line 78:
```suggestion
				std::array<double,3> llh_0 = {origin.latitude_deg, origin.longitude_deg, origin.altitude_m};
```

modules/ObjectControl/src/objectlistener.cpp
```diff
@@ -245,4 +256,21 @@ atos_interfaces::msg::Monitor createROSMessage(const MonitorMessage& monrMessage
 	msg.acceleration.accel.angular.y = 0;
 	msg.acceleration.accel.angular.z = 0;
 	return msg;
+}
+
+sensor_msgs::msg::NavSatFix createNavSatFixMessage(const struct timeval& tv, double origin[3], const atos_interfaces::msg::Monitor &monr) {
```
LukasWikander, line 261: Good to know that the compiler treats
```C++
void myFun(double var[123]) {
    ...
}
```
the same as
```C++
void myFun(double* var) {
    ...
}
```
which can produce annoying errors.

```suggestion
sensor_msgs::msg::NavSatFix createNavSatFixMessage(const struct timeval& tv, std::array<double,3> origin, const atos_interfaces::msg::Monitor &monr) {
```

LukasWikander (approved):
Just some minor stuff

LukasWikander:
Also, don't forget the license headers in all the new files @victorjarlow !

victorjarlow:
> Also, don't forget the license headers in all the new files @victorjarlow !

 Thanks for the reminder :) takes some getting used to

modules/EsminiAdapter/src/esminiadapter.cpp
victorjarlow: needed for UtilGetOscDirectoryPath, this can be addressed in a future pr where we break out util into e.g. fileutils etc.

modules/ObjectControl/src/objectlistener.cpp
victorjarlow: Missed this one, thanks :+1:

PR #531, opened 2023-03-06
Updated scenario at flex zone for FFI demo

LukasWikander:


scripts/openscenario/ffi_scenario.py
```diff
@@ -33,49 +33,49 @@
 mondeo_slow_speed = 10 / 3.6 # m/s
 mondeo_acceleration = 0.51 # m/s^2
 mondeo_retardation = 2.17 # m/s^2
-mondeo_O1_point = xosc.LanePosition(83.5103988647461, 0, -1, 3)
+mondeo_O1_point = xosc.LanePosition(15, 0, -1, 2)
```
victorjarlow, line 36: might improve readability to call it mondeo_starting_point

victorjarlow: Also creating variables for the lanes / roads e.g. LEFT_LANE = -1

victorjarlow (approved):

scripts/openscenario/ffi_scenario.py
LukasWikander: The lane IDs are not consistent since it is not fixed in what "direction" a road is pointing

victorjarlow: Ok I see, too bad!

PR #532, opened 2023-03-07
Added atos_interfaces as a submodule

LukasWikander:
To make the most use of this, you need to `ln -s /path/to/git/ATOS/atos_interfaces atos_interfaces` in your `/path/to/ros_ws/src` dir


victorjarlow (approved):
lgtm

PR #533, opened 2023-03-07
Made a CMakeLists with build step npm install

LukasWikander:
From what I can tell this setup is correct. However, when running `npm rebuild` outside of CMake with the correct environment I get a strange permission error that I believe is not related to the changes in this PR:
image


victorjarlow (approved):
works on my machine ™

victorjarlow:
> works on my machine ™

This generates atos_interfaces in install/atos/etc/simplecontrol/node_modules/rclnodejs/generated/ for me

PR #540, opened 2023-03-11
make sure esminiAdapter stops on abort / shuts down on exit

victorjarlow:
small fix to make message callbacks static. If callbacks are not static then the message event has no effect

### DIFF
--- a/modules/EsminiAdapter/src/esminiadapter.cpp
+++ b/modules/EsminiAdapter/src/esminiadapter.cpp
@@ -50,6 +50,8 @@ EsminiAdapter::EsminiAdapter() : Module(moduleName),
 	v2xPub(*this),
 	initSub(*this, &EsminiAdapter::onStaticInitMessage),
 	startSub(*this, &EsminiAdapter::onStaticStartMessage),
+	abortSub(*this, &EsminiAdapter::onStaticAbortMessage),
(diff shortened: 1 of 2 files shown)


modules/EsminiAdapter/src/esminiadapter.cpp
```diff
@@ -123,18 +127,11 @@ void EsminiAdapter::onConnectedObjectIdsMessage(const ConnectedObjectIds::messag
 
 //! Message queue callbacks
 
-void EsminiAdapter::onAbortMessage(const Abort::message_type::SharedPtr) {
+void EsminiAdapter::onStaticAbortMessage(const Abort::message_type::SharedPtr) {
```
LukasWikander, line 130: Shouldn't the name of this method be "closeScenarioEngine"? Right now it is named based on what will call it, not what it does.

LukasWikander (approved):
Isn't there some other way of completing a test than with abort that should be handled?

victorjarlow:
> Isn't there some other way of completing a test than with abort that should be handled?

Right now esminiAdapter isnt too aware of the state of the current test. ScenarioEngine is started in onStartMessage() and stopped when onInitMessage() is called. I think a future PR can make sure that esminiAdapter detects the end of the scenario and stops itself.

PR #543, opened 2023-03-13
Feature pointcloud

samuelthoren:
Added a pointcloud publisher in order to visualise our pointclouds in foxglove. In order to run this you will need the following:
- Point Cloud Library installed (`sudo apt install libpcl-dev`)
- Add a directory in ATOS called `pointclouds`
- A `.pcd`-file
- Add the following in `params.yaml`:
```pointcloud_publisher:
    ros__parameters:
      pointcloud_file: "my_pointcloud_file.pcd"
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ Prerequisites: C/C++ compiler, CMake (minimum version 3.10.2)
 ## <a name="dependencies"></a> Dependencies & external libraries
 In order to build ATOS, dependencies and exernal libraries need to be installed. First install the necessary development packages:
 ```
-sudo apt install libsystemd-dev libprotobuf-dev protobuf-compiler libeigen3-dev ros-foxy-paho-mqtt-c nlohmann-json3-dev
+sudo apt install libsystemd-dev libprotobuf-dev protobuf-compiler libeigen3-dev ros-foxy-paho-mqtt-c nlohmann-json3-dev libpcl-dev
 ```
 
 Then, the following external libraries need to be installed:
@@ -100,7 +100,7 @@ echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/ros-a
 Install ros foxy for desktop and colcon
 ```
 sudo apt update
-sudo apt install ros-foxy-desktop python3-colcon-common-extensions ros-foxy-nav-msgs ros-foxy-geographic-msgs ros-foxy-foxglove-msgs ros-foxy-sensor-msgs  ros-foxy-rosbridge-suite
+sudo apt install ros-foxy-desktop python3-colcon-common-extensions ros-foxy-nav-msgs ros-foxy-geographic-msgs ros-foxy-foxglove-msgs ros-foxy-sensor-msgs  ros-foxy-rosbridge-suite ros-foxy-pcl-conversions
 ```
 
 source the setup script:
--- a/common/roschannels/pointcloudchannel.hpp
+++ b/common/roschannels/pointcloudchannel.hpp
@@ -1,0 +1,28 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+#pragma once
+
+#include "roschannel.hpp"
+#include "sensor_msgs/msg/point_cloud2.hpp"
+
+namespace ROSChannels {
+  namespace Pointcloud {
+    const std::string topicName = "pointcloud";
+    using message_type = sensor_msgs::msg::PointCloud2;
+    
+    class Pub : public BasePub<message_type> {
+    public:
+        Pub(rclcpp::Node& node) :
+          BasePub<message_type>(node, "pointcloud/" + topicName, rclcpp::QoS(rclcpp::KeepLast(1)).transient_local()) {}
+    };
+
+    class Sub : public BaseSub<message_type> {
+    public:
+      Sub(rclcpp::Node& node, std::function<void(const message_type::SharedPtr)> callback) : 
+          BaseSub<message_type>(node, "pointcloud/" + topicName, callback, rclcpp::QoS(rclcpp::KeepLast(1)).transient_local()) {}
+    };
+  }
+}
--- a/common/roschannels/pointcloudchannel.hpp
+++ b/common/roschannels/pointcloudchannel.hpp
@@ -15,14 +15,16 @@ namespace ROSChannels {
     
     class Pub : public BasePub<message_type> {
     public:
-        Pub(rclcpp::Node& node) :
-          BasePub<message_type>(node, "pointcloud/" + topicName, rclcpp::QoS(rclcpp::KeepLast(1)).transient_local()) {}
+        const std::string fileName;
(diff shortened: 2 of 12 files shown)


modules/PointcloudPublisher/inc/pointcloudpublisher.hpp
```diff
@@ -0,0 +1,41 @@
...
+  private:
+    static inline std::string const moduleName = "pointcloud_publisher";
+    ROSChannels::Init::Sub initSub;
+		ROSChannels::Pointcloud::Pub pointcloudPub;
+
+    std::string pointcloudFile;
+    sensor_msgs::msg::PointCloud2 msg;
+    std::shared_ptr<pcl::PointCloud<pcl::PointXYZRGB>> pointcloud;
+
+    // std::shared_ptr<rclcpp::Publisher<sensor_msgs::msg::PointCloud2>> publisher;
```
LukasWikander, line 32:
```suggestion
```

modules/PointcloudPublisher/inc/pointcloudpublisher.hpp
```diff
@@ -0,0 +1,41 @@
...
+    std::shared_ptr<rclcpp::TimerBase> timer;
+
+		void onInitMessage(const ROSChannels::Init::message_type::SharedPtr) override;
```
LukasWikander, line 35:
```suggestion
    void onInitMessage(const ROSChannels::Init::message_type::SharedPtr) override;
```

modules/PointcloudPublisher/src/pointcloudpublisher.cpp
```diff
@@ -0,0 +1,90 @@
...
+  
+}
+
+
+/**
+ * @brief Publish pointcloud on init message
+ * 
+ */
+void PointcloudPublisher::onInitMessage(const ROSChannels::Init::message_type::SharedPtr) {
+  pointcloudPub.publish(msg);
```
LukasWikander, line 86: Also need to call initialize, right? Otherwise it is not possible to change the configuration once ATOS is running.

LukasWikander (changes requested):
Crashing if the file does not exist is not a good solution. Init is the way we signal to ATOS that a new config is available so simply printing that it does not exist is an OK way to handle it. That gives the user the ability to upload the correct file and press Init again, and not have to restart ATOS

common/roschannels/pointcloudchannel.hpp
```diff
@@ -0,0 +1,28 @@
...
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+#pragma once
+
+#include "roschannel.hpp"
+#include "sensor_msgs/msg/point_cloud2.hpp"
+
+namespace ROSChannels {
+  namespace Pointcloud {
+    const std::string topicName = "pointcloud";
```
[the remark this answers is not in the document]
samuelthoren: Yes, this was still in progress when you reviewed it. I have added support for multiple pointclouds, so the topic is now /atos/pointcloud/pointcloudName

README.md
```diff
@@ -32,7 +32,7 @@ Prerequisites: C/C++ compiler, CMake (minimum version 3.10.2)
 ## <a name="dependencies"></a> Dependencies & external libraries
```
LukasWikander, line 32: Maybe add to this file, where to place pointcloud files, how to specify their coordinate systems etc.

conf/params.yaml
```diff
@@ -31,4 +31,7 @@ atos:
       quality_of_service: ""
   trajectorylet_streamer:
     ros__parameters:
-      chunk_duration: 2.0
\ No newline at end of file
+      chunk_duration: 2.0
+  pointcloud_publisher:
+    ros__parameters:
+      pointcloud_file: [""]
```
LukasWikander, line 37:
```suggestion
      pointcloud_files: [""]
```
... with corresponding changes in code

modules/PointcloudPublisher/src/pointcloudpublisher.cpp
```diff
@@ -0,0 +1,119 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+
+#include "pointcloudpublisher.hpp"
+#include <pcl_conversions/pcl_conversions.h>
+#include <pcl/filters/voxel_grid.h>
```
LukasWikander, line 9:
```suggestion
```

modules/PointcloudPublisher/inc/pointcloudpublisher.hpp
```diff
@@ -0,0 +1,37 @@
...
+
+    std::vector<std::string> pointcloudFiles;
+    std::map<std::string, std::shared_ptr<pcl::PointCloud<pcl::PointXYZRGB>>> pointclouds;
+
+    void onInitMessage(const ROSChannels::Init::message_type::SharedPtr) override;
+    void initialize();
+    void getPointcloudFiles();
+    void loadPointClouds();
+    void createPublishers();
+    std::string getPublisherTopicName(const std::string& path);
```
LukasWikander, line 35:
```suggestion
    std::string getPublisherTopicName(const std::string& path) const;
```

modules/PointcloudPublisher/inc/pointcloudpublisher.hpp
```diff
@@ -0,0 +1,37 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+#pragma once
+
+#include "module.hpp"
+#include "roschannels/pointcloudchannel.hpp"
+#include "roschannels/commandchannels.hpp"
+#include <sensor_msgs/msg/point_cloud2.hpp>
+#include <pcl/io/pcd_io.h>
```
LukasWikander, line 12: Is this include needed in the header?

modules/PointcloudPublisher/inc/pointcloudpublisher.hpp
```diff
@@ -0,0 +1,37 @@
...
+    static inline std::string const moduleName = "pointcloud_publisher";
+    ROSChannels::Init::Sub initSub;
+    std::map<std::string, std::shared_ptr<ROSChannels::Pointcloud::Pub>> pointcloudPubs;
+
+    std::vector<std::string> pointcloudFiles;
+    std::map<std::string, std::shared_ptr<pcl::PointCloud<pcl::PointXYZRGB>>> pointclouds;
+
+    void onInitMessage(const ROSChannels::Init::message_type::SharedPtr) override;
+    void initialize();
+    void getPointcloudFiles();
```
LukasWikander, line 32: I would expect this method to return the member `pointclouds` or similar. Is there a better name that shows this reads from parameters?

LukasWikander (approved):
Minor stuff. Only major thing is that you're using spaces and not tabs :scream:

README.md
samuelthoren: I think this is a good thing to have, but it could be done in another PR where we explain all directories (`osc`, `odr`, `pointclouds`, etc.) and what should be placed in them in order to get ATOS running as well as explaining `params.yaml`

PR #548, opened 2023-03-21
User manual for ATOS

samuelthoren:
Hi! I started on a user manual for ATOS explaining which settings you can do, and how to get started with the graphical user interfaces. Right now I have not written a lot, but I would like to get some feedback on the structure of it and what should be added/removed/changed before I get in too deep, so don't approve this PR yet ;). I have also started a section where the ROS parameters are described, and would be thankful if I could get some help on writing those since there are some modules I have not worked very much with. I have included an example for the `PointcloudPublisher` how this could look.

In the future this manual maybe shouldn't be located in the `README`, but for now I just wanted to get something down in text and then we can move it later.

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,17 +176,17 @@ After starting ATOS the first time, you will find the test directory in `~/.asta
   - Explanation here: 
   - Supported file types:
 - objects
-  - Explanation here: 
-  - Supported file types:
+  - Explanation here: Directory containing all objects that should be used in a test.
+  - Supported file types: `.opro`
 - odr
-  - Explanation here: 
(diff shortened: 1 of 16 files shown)


LukasWikander:
We should not remove the basic build instructions from the README. This page serves as our intro page for developers. We want to make it clear that it is simple to get started with only a few steps, not direct them to an in-depth guide right away.

samuelthoren:
I will close this PR since all of these commits has been moved here instead: https://github.com/RI-SE/ATOS/tree/feature_read_the_docs

PR #554, opened 2023-03-24
Add github action for version tagging

Robert108:
Default version to bump is "patch" but can be overridden by having a "hashtag:" (with an actual hashtag symbol) and then specifying minor or major in the PR commit message. 

See docs here https://github.com/anothrNick/github-tag-action


victorjarlow:
Is it possible to log the currently running software version on startup? á la safety cage https://github.com/RI-SE/SafetyCage/tree/main/safeguard/src @LukasWikander how did you achieve project version printing?

LukasWikander:
It needs to be built into the code to achieve that effect. I agree it would be useful to e.g., be able to check logs or know one is not accidentally running the wrong binary.

launch/launch_utils/launch_base.py
```diff
@@ -5,12 +5,17 @@
...
+import rclpy.logging as logging
 
 def get_base_nodes():
     files = validate_files()
     atos_conf_dir = os.path.join(os.path.expanduser('~'), '.astazero', 'ATOS')
     atos_install_dir = get_package_prefix('atos')
 
+    # Print ATOS version
+    version_file = open(atos_install_dir / Path(".VERSION"), 'r')
+    logging.get_logger('launch').info("ATOS version: " + version_file.read())
```
LukasWikander, line 17:
```suggestion
def print_version():
    version_file = open(atos_install_dir / Path(".VERSION"), 'r')
    logging.get_logger('launch').info("ATOS version: " + version_file.read())
```

LukasWikander: Also need to call this right after `get_base_nodes()` is called

LukasWikander (approved):
Minor stuff

PR #556, opened 2023-03-24
according to clang++, v.SizeAtCompileTime is not an integral constant…

victorjarlow:
… expression, and therefore assert should be used instead of static_assert


LukasWikander:
What does the && with the string do?

LukasWikander (approved):

PR #557, opened 2023-03-24
Feature no pthread cancel

victorjarlow:
Going through old logs, it seems that the following two lines appear together leading up to a crash of object_control:

1649083846.0016840 [object_control-3] failed to take global rclcpp logging mutex
1649083846.0018322 [object_control-3] FATAL: exception not rethrown

FATAL: exception not rethrown can be caused by using pthread_cancel which we utilized. 
My hypothesis is that a thread might have been cancelled while holding the global logging mutex, causing some other thread within the object_control process to be unable to take the global rclcpp logging mutex.

So hopefully by getting rid of pthread_cancel, this bug ceases to be. Would be nice to verify that this does not occur anymore somehow.

### DIFF
--- a/modules/ObjectControl/inc/testobject.hpp
+++ b/modules/ObjectControl/inc/testobject.hpp
@@ -86,7 +86,9 @@ public:
 	ObjectConnection(rclcpp::Logger log)
 		: cmd(SOCK_STREAM, log),
 		mntr(SOCK_DGRAM, log),
-		Loggable(log) {}
+		Loggable(log) {
+			pipe(interruptionPipeFds);
+		}
 
 	bool isValid() const;
 	bool isConnected() const;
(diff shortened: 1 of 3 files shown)


modules/ObjectControl/inc/testobject.hpp
```diff
@@ -86,14 +86,19 @@ class ObjectConnection : public Loggable {
 	ObjectConnection(rclcpp::Logger log)
 		: cmd(SOCK_STREAM, log),
 		mntr(SOCK_DGRAM, log),
-		Loggable(log) {}
+		Loggable(log) {
+			pipe(interruptionPipeFds);
```
LukasWikander, line 90: Does this need to be "un-piped" at destruction?

LukasWikander (approved):

PR #563, opened 2023-03-30
Headline update

viktorjo:
To understand the meaning of ATOS, I suggest we write out the name in the GitHub readme.


samuelthoren:
We have a large branch that will be merged soon, where we are currently writing all documentation. I added your suggestion there instead.

https://github.com/RI-SE/ATOS/blob/feature_read_the_docs/README.md

PR #568, opened 2023-05-02
Feature simplify installation

Robert108:
This PR updates and fixes the branch feature_DockerImage by @eriadam36 to work with the latest ATOS version. It also adds an installation script that should help a lot when installing ATOS from source.


Robert108:
This will need some further testing on a clean installation of Ubuntu 20.04 and 22.04.
Right now the docker images will not be able connect to local instances of objects on ports 53241 and 53240. Should likely be able to fix this with some extra network configuration in docker.

CMakeLists.txt
```diff
@@ -9,7 +9,14 @@ set(CMAKE_CXX_STANDARD 11)
...
-set( CMAKE_EXPORT_COMPILE_COMMANDS ON )
+
+#Set preprocessor macros depending on used ros version
+if ($ENV{ROS_DISTRO} STREQUAL "foxy")
+  add_definitions(-DROS_FOXY)
+elseif($ENV{ROS_DISTRO} STREQUAL "humble")
+  add_definitions(-DROS_HUMBLE)
+endif()
+
 project(atos VERSION 1.0.0)
```
sepast, line 20: Maybe this is version 1.0.1?

common/module.cpp
```diff
@@ -48,7 +48,7 @@ void Module::tryHandleMessage(
 bool Module::requestDataDictInitialization(int maxRetries) {
     SetBool::Response::SharedPtr response;
     using namespace std::chrono_literals;
-    auto successful = nShotServiceRequest<SetBool>(maxRetries, 1s, ServiceNames::initDataDict, response);
+    auto successful = nShotServiceRequest<SetBool>(maxRetries, 2s, ServiceNames::initDataDict, response);
```
sepast: 2 s seems to be a long period, has this timeout been a problem in the passed?

common/module.hpp
```diff
@@ -79,7 +79,7 @@ class Module : public rclcpp::Node {
 								 const std::string& topic,
 								 const rclcpp::Logger& logger);
 								 
-	bool requestDataDictInitialization(int maxRetries = 3);
+	bool requestDataDictInitialization(int maxRetries = 5);
```
sepast: Relates to the previous comment, has it been a problem with initialization with DataDictionary in the passed?

modules/ATOSBase/src/ATOSbase.cpp
```diff
@@ -18,10 +18,10 @@ ATOSBase::ATOSBase()
...
 	exitSub(*this, std::bind(&ATOSBase::onExitMessage, this, _1))
 {
-	declare_parameter("test_origin_latitude");
-	declare_parameter("test_origin_longitude");
-	declare_parameter("test_origin_altitude");
-	declare_parameter("test_origin_rot");
+	declare_parameter("test_origin_latitude", 57.77744364201314);
+	declare_parameter("test_origin_longitude", 12.7809807263866);
+	declare_parameter("test_origin_altitude",193.114);
+	declare_parameter("test_origin_rot",0.0);
```
sepast, line 25: test origin is changed on multiple places - is this another one?

modules/MQTTBridge/src/mqttbridge.cpp
```diff
@@ -7,19 +7,20 @@
...
 MqttBridge::MqttBridge() : Module(MqttBridge::moduleName),
 						   v2xMsgSub(*this, std::bind(&MqttBridge::onV2xMsg, this, _1))
 {
-	declare_parameter("broker_ip");
-	declare_parameter("pub_client_id");
-	declare_parameter("username");
-	declare_parameter("password");
-	declare_parameter("topic");
-	declare_parameter("quality_of_service");
+	declare_parameter("broker_ip","192.71.100.198");
```
sepast: Hard coding an IP-address?

sepast (approved):

modules/ATOSBase/src/ATOSbase.cpp
konglobemeralt: This is just a declaration of the parameter for ROS, it needs to have a default value, perhaps something better could be choosen, null island perhaps?

modules/MQTTBridge/src/mqttbridge.cpp
konglobemeralt: Same here, a value is required when a parameter is declared. Could be localhost or something to clarify that it is something to be changed. On the other hand for us this is the broker adress and we should not use another.

CMakeLists.txt
Robert108: Hmmm I wonder if should have version tagging at all here in the CMake file now when we have it on git as well.
I any case we should update the version number automatically here (to be the same as the git tag) instead of hard coding it here.

modules/ATOSBase/src/ATOSbase.cpp
Robert108: Changed it to null island instead

konglobemeralt (approved):

modules/ATOSBase/src/ATOSbase.cpp
eriadam36: not sure if ros will like that on a clean install. 
I think I tried it and it complains with the types, it wanted to know if it was floats int:s etc.... don't know why

eriadam36:
Did you end up adding the atos_interface as a submodule or not?
It was my quick and dirty way to solve the installation for docker. Don't know if it the best of ideas :smile:

PR #571, opened 2023-05-15
Path file parsing 

dandeandean:
We created path.hpp/cpp in drone control module. You can read from a path (.xml) file as well as manipulate the read data by adding an offset


dandeandean:
wrong branch

PR #578, opened 2023-06-05
Feature drone control

victorjarlow:
Module for parsing ABD .path files and creating trajectories for filming drones to follow.

This module introduces a new source for trajectories, therefore the change was made to move IP addresses to the .opro files again. 

If this module is enabled, EsminiAdapter should not be enabled (at least it should not provide trajectories to OBC). 

Idea for future PR: split esmini adapter into two modules, one that provides trajectories, and one that acts as a simulation backend.


SebastianSvanland:
@victorjarlow What is the status here?

PR #579, opened 2023-06-05
Feature ros node test objects final2023

victorjarlow:
New features:
* Splits up testobject.cpp into multiple files, each containing a class.
* make each TestObject a ros2 node
* communicate the state change of a TestObjects via a ros topic (atos_interfaces::msg::StateChange)

### DIFF
--- a/common/roschannels/monitorchannel.hpp
+++ b/common/roschannels/monitorchannel.hpp
@@ -98,7 +98,7 @@ namespace ROSChannels {
             ObjectMonitorType outdata;
             outdata.timestamp.tv_sec = indata.atos_header.header.stamp.sec;
             outdata.timestamp.tv_usec = indata.atos_header.header.stamp.nanosec / 1000;
-            //TODO: Add state..
+            outdata.state = static_cast<ObjectStateType>(indata.object_state.state);
             outdata.position.isPositionValid = true;
             outdata.position.isXcoordValid = true;
             outdata.position.isYcoordValid = true;
--- a/common/roschannels/monitorchannel.hpp
+++ b/common/roschannels/monitorchannel.hpp
@@ -30,6 +30,12 @@ namespace ROSChannels {
                 BasePub<message_type>(node, "object_" + std::to_string(id) + "/" + topicName, qos) {}
         };
 
+        class AnchorPub : public BasePub<message_type> {
+        public:
+            AnchorPub(rclcpp::Node& node, const rclcpp::QoS& qos = defaultQoS) :
+                BasePub<message_type>(node, "object_anchor/" + topicName, qos) {}
+        };
+
         class Sub : public BaseSub<message_type> {
         public:
             const uint32_t objectId;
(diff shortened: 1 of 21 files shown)


modules/ObjectControl/src/objectconnection.cpp
```diff
@@ -0,0 +1,79 @@
...
+		}
+		fd_set fds;
+		FD_ZERO(&fds);
+		FD_SET(interruptionPipeFds[0], &fds);
+		FD_SET(mntr.socket, &fds);
+		FD_SET(cmd.socket, &fds);
+		auto result = select(std::max({mntr.socket,cmd.socket,interruptionPipeFds[0]})+1,
+							 &fds, nullptr, nullptr, nullptr);
+		if (result < 0) {
+			throw std::runtime_error(std::string("Failed socket operation (select: ") + strerror(errno) + ")"); // TODO clearer
```
sepast, line 58: Is this something that needs to get fixed in this PR?

sepast (approved):
Looks ok
This is a large PR so it's hard to if it works or not.

modules/ObjectControl/src/objectconnection.cpp
victorjarlow: Are you referring to the TODO?

modules/ObjectControl/src/relativetestobject.cpp
```diff
@@ -0,0 +1,129 @@
...
+ * \param anchor Monitor data of anchor to which point is relative
+ * \return Transformed monitor data
+ */
+MonitorMessage RelativeTestObject::readMonitorMessage() {
+    MonitorMessage retval;
+    this->comms.mntr >> retval;
+    lastMonitorTime = clock::now();
+    updateMonitor(retval);
+    // transform the monitor data relative to anchor
+	transformCoordinate(retval.second,lastAnchorMonr,true);
```
konglobemeralt, line 46:
```suggestion
    transformCoordinate(retval.second,lastAnchorMonr,true);
```

konglobemeralt: Small formatting fix

modules/ObjectControl/src/relativetestobject.cpp
```diff
@@ -0,0 +1,129 @@
...
+	ObjectMonitorType retval = point;
+	struct timeval tvdiff = {0, 0};
+	std::chrono::milliseconds diff;
+	if (point.isTimestampValid && anchor.isTimestampValid) {
+		timersub(&point.timestamp, &anchor.timestamp, &tvdiff);
+	}
+	from_timeval(tvdiff, diff);
+
+	AngleAxis anchorToGlobal(anchor.position.heading_rad, Vector3d::UnitZ());
+    AngleAxis pointToGlobal(point.position.heading_rad, Vector3d::UnitZ());
```
konglobemeralt:
```suggestion
	AngleAxis pointToGlobal(point.position.heading_rad, Vector3d::UnitZ());
```
Same here

common/roschannels/monitorchannel.hpp
```diff
@@ -22,12 +30,99 @@ namespace ROSChannels {
...
+            outdata.acceleration.accel.angular.y = 0;
+            outdata.acceleration.accel.angular.z = 0;
+            return outdata;
+        }
+
+        static ObjectMonitorType toISOMonr(message_type& indata){
+            ObjectMonitorType outdata;
+            outdata.timestamp.tv_sec = indata.atos_header.header.stamp.sec;
+            outdata.timestamp.tv_usec = indata.atos_header.header.stamp.nanosec / 1000;
+            //TODO: Add state..
```
konglobemeralt: Will state be added in an additional PR? There is a OBJECT_STATE_UNKNOWN that could be used instead of leaving garbage data in the state slot before this is handled.

common/roschannels/monitorchannel.hpp
```diff
@@ -22,12 +30,99 @@ namespace ROSChannels {
...
+
+            // Set frame ids
+            outdata.atos_header.header.frame_id = "map"; // TODO
+            outdata.pose.header.frame_id = "map"; // TODO
+            outdata.velocity.header.frame_id = "map"; // TODO vehicle local
+            outdata.acceleration.header.frame_id = "map"; // TODO vehicle local
+
+            outdata.atos_header.object_id = txid;
+            outdata.object_state.state = indata.state;
+            if (indata.position.isPositionValid) {
```
konglobemeralt, line 72: Is isPosition valid only checking if x and y is valid?

common/roschannels/monitorchannel.hpp
victorjarlow: I just forgot to do this, thanks!

common/roschannels/monitorchannel.hpp
victorjarlow: Can you clarify what the issue is?

konglobemeralt: I was just wondering about the isValid check on the Z coordinate a few lines down on 75. I am assuming we are still allowing z to be ignored. I am just thinking if this could be an issue for drones. But then again, I since this is the monitor message I assume this is due to iso22133.

common/roschannels/objstatechangechannel.hpp
```diff
@@ -0,0 +1,27 @@
...
+#pragma once
+
+#include "roschannel.hpp"
+#include "atos_interfaces/msg/object_state_change.hpp"
+
+namespace ROSChannels {
+    namespace ObjectStateChange {
+        const std::string topicName = "object_state_change";
+        using message_type = atos_interfaces::msg::ObjectStateChange;
+        const rclcpp::QoS defaultQoS = rclcpp::QoS(rclcpp::KeepAll());
```
konglobemeralt, line 15: A bit nitpicky perhaps but why is this called defaultQos? Can a node be expected to change its qos to keep(1) or something else? 

Also is this is more of a question, but is there a specific reason to use keep all instead of keep N? Does ROS handle this if it gets to large? I assume might not be that big of a risk considering that most runs do not include that many state changes.

samuelthoren:
I have just started testing, but when I press `Init` I get an extra node called `/atos/object_control` which does not seem correct.

To reproduce:
- Start ATOS, run `ros2 node list`
- Press `Init`
- Run `ros2 node list` again and there will be two nodes with the same name

If I press `Disconnect` and `Init` again however, there are still two nodes, so I am not sure why this isn't creating a third node.

After some more testing it seems that this depends on the number of objects that gets initialised, for instance for two objects there are three nodes that are called `/atos/object_control`, it seems that the correct name for the nodes for the objects are never set.

victorjarlow:
> I have just started testing, but when I press `Init` I get an extra node called `/atos/object_control` which does not seem correct.
> 
> To reproduce:
> 
> * Start ATOS, run `ros2 node list`
> * Press `Init`
> * Run `ros2 node list` again and there will be two nodes with the same name
> 
> If I press `Disconnect` and `Init` again however, there are still two nodes, so I am not sure why this isn't creating a third node.
> 
> After some more testing it seems that this depends on the number of objects that gets initialised, for instance for two objects there are three nodes that are called `/atos/object_control`, it seems that the correct name for the nodes for the objects are never set.

Thanks for spotting this!
I've addressed it by removing the object_control name from the launch file and adding the nodes to the executor.

common/roschannels/objstatechangechannel.hpp
victorjarlow: > A bit nitpicky perhaps but why is this called defaultQos?

The variable is named defaultQoS because it is the default QoS of the channel. 
> Can a node be expected to change its qos to keep(1) or something else?

When constructing a ROSChannels::ObjectStateChange::Pub or ::Sub object, another value for QoS, any value can be input as a constructor-parameter, but we have not implemented changing this value dynamically at runtime. 

>Also is this is more of a question, but is there a specific reason to use keep all instead of keep N?

This is the size of the callback queue for subscribers of this type. If there are system-wide delays there might be useful to have a complete history of state changes. Example with delays and KeepLast(2): off -> initialized -> disarmed -> armed -> disarmed. In this case, OBC will get a state change from off to armed which will be illegal. With the full history the state changes will be correct. I don't think that this is a major issue though.. 

>Does ROS handle this if it gets to large?

This is subject to the resources of the underlying DDS vendor (https://design.ros2.org/articles/qos.html).

>I assume might not be that big of a risk considering that most runs do not include that many state changes.
Yes I agree

common/roschannels/monitorchannel.hpp
victorjarlow: added isvalid checks for x and y too

PR #582, opened 2023-07-19
Feature add doxygen docs

Robert108:


samuelthoren (approved):
Approving since Victor can't approve it

PR #586, opened 2023-09-05
Update ATOS iso22133 submodule ref

Robert108:
This PR updates the iso22133 submodule which introduces new signals from the ISO22133 specification. 
It also update the node version in the installation script for ROS foxy and removes the package-lock.json file which is a common source for build failures since we support multiple node/ROS versions.


sepast (approved):
I've tested this and it works fine to execute `colcon build`

PR #589, opened 2023-09-21
Integration testing

samuelthoren:
Integration testing for ATOS. The idea is that one can select which integration tests to run in `params.yaml` and then `IntegrationTestingHandler` reads which tests to be executed and runs them. `IntegrationTesting` is a base class which all integration tests inherits from, making it easy to add future integration tests. In this PR I have implemented `ScenarioExecution`, which is an integration test in order to test if a scenario can be executed. It sends the commands: init, connect, arm, start, and then checks if all states are correct. It also checks if the object follows the trajectory correctly.

In order to run the test, start this version of the ISO-object: https://github.com/RI-SE/isoObject/pull/12, and then do `ros2 launch atos launch_integration_testing.py`. ATOS will then run and execute the selected integration tests, and then output if the integration test passed or not. The output may look like this:
image

This should work with any scenario, since the ISO-object automatically finds the start of the trajectory in the scenario.


Robert108:
Nice work!
As discussed earlier we should try to fix automatic setup for the tests, this will help both with CI and make sure that users actually perform the tests. 
Using the dockerized isoObject could work well together with docker compose to set up a test environment. 
Fixing parameter handling and configuration should also be part of this PR.
I'm happy to support in both these tasks.

common/module.hpp
```diff
@@ -103,6 +103,34 @@ class Module : public rclcpp::Node {
...
+		if (rclcpp::spin_until_future_complete(get_node_base_interface(), promise, timeout) ==
+			rclcpp::FutureReturnCode::SUCCESS) {
+			response = promise.get();
+			return true;
+		} else {
+			RCLCPP_ERROR(get_logger(), "Failed to call service %s", client->get_service_name());
+			return false;
+		}
+	}
+
```
victorjarlow: This function and and the above function can be merged?
Use the following parameter:
`std::shared_ptr<typename Srv::Request> &request = std::make_shared<typename Srv::Request>(),`

modules/IntegrationTesting/src/scenarioexecution.cpp
```diff
@@ -0,0 +1,149 @@
...
+	}
+	return trajectory;
+}
+
+
+/**
+ * @brief Check that the object ends up near the end point of the trajectory.
+ * 
+ */
+void ScenarioExecution::checkTrajectory() {
```
victorjarlow: Could this check be also along the trajectories? Maybe not for this PR but as a future improvement

modules/IntegrationTesting/src/scenarioexecution.cpp
```diff
@@ -0,0 +1,149 @@
...
+ * arm the object, and then start it. It will check that all states are correct, and that the object follows
+ * the trajectory.
+ * 
+ */
+void ScenarioExecution::runIntegrationTest() {
+	RCLCPP_INFO(get_logger(), "Running scenario");
+	auto msg = std_msgs::msg::Empty();
+
+	initPub->publish(msg);
+	std::this_thread::sleep_for(std::chrono::seconds(1)); // sleep to allow the object to initialize
```
victorjarlow:
```suggestion
	std::this_thread::sleep_for(1s); // sleep to allow the object to initialize
```

victorjarlow: same below

victorjarlow (commented):
Good job! Can this be connected to the gtest method in https://github.com/RI-SE/ATOS/pull/599 somehow? It seems that the printout is basically a gtest report comparing actual to expected state. Maybe some of the boilerplate introduced in this PR could be handled by gtest?

modules/IntegrationTesting/src/scenarioexecution.cpp
samuelthoren: Yes, I think it's a good idea to extend this in the future. But I think it's good enough for now.

samuelthoren:
Robert and I added support for running one integration test with one ISO-object. This can be run both offline and as a GitHub action.

common/module.hpp
```diff
@@ -82,19 +82,24 @@ class Module : public rclcpp::Node {
...
-	/*! \brief This helper function is performs a service call given a client and yields a response.
-	*  \tparam Srv The name of the service to request.
-	*  \param timeout The timeout for the service call.
-	*  \param client The client to use to request the service.
-	*  \param response The response of the service.
-	* 	\return The response of the service.
-	*/
+
+	/**
+	 * @brief This helper function is performs a service call given a client and yields a response. This
```
SebastianSvanland: Typo

```suggestion
	 * @brief This helper function performs a service call given a client and yields a response. This
```

docs/Contributing/Testing/testing.md
```diff
@@ -20,5 +20,20 @@ colcon and cmake switches if test should be built with the BUILD_TESTING flag. T
 colcon build --cmake-args -DBUILD_TESTING=ON
 ```
 
+## Integration testing
+The docker compose file docker-compose-test.yml is used to run integration tests towards one or more simulated ISO22133 test objects. These test are used to "smoke test" ATOS and makes sure that no changes 
+breaks any core functionality. You can start these test with the command
```
SebastianSvanland, line 25: Any reason for these lines to be split into two in the middle of a sentence? Or just an oopsie?

docs/Usage/Modules/IntegrationTesting.md
```diff
@@ -0,0 +1,19 @@
+# IntegrationTesting
+
+## About the module
+`IntegrationTesting` is used for integration tests. The module launches an integration testing handler, which reads from `params.yaml` which integration tests we want to execute. A node for each integration
+test is then launched, and the node performs the specified integration tests. An integration tests can for instance be if an object reports all states correctly during a run, or if it follows a trajectory
+correctly. This module can be run both offline and through GitHub Actions.
```
SebastianSvanland, line 6: Why are there sentences that have been split up between multiple lines?

docs/Usage/Modules/IntegrationTesting.md
```diff
@@ -0,0 +1,19 @@
...
+
+!!! note
+
+    You can read more about integration testing in the section [Testing](../../Contributing/Testing/testing.md).
```
SebastianSvanland, line 10: The link to the testing.md isn't working since it's written in a code section for some reason. Not sure if it helps to just remove the blank spaces before
```suggestion
You can read more about integration testing in the section [Testing](../../Contributing/Testing/testing.md).
```

modules/IntegrationTesting/CMakeLists.txt
```diff
@@ -0,0 +1,60 @@
...
+	${CMAKE_CURRENT_SOURCE_DIR}/src/main.cpp
+	${CMAKE_CURRENT_SOURCE_DIR}/src/integrationtestinghandler.cpp
+	${CMAKE_CURRENT_SOURCE_DIR}/src/integrationtestingfactory.cpp
+	${CMAKE_CURRENT_SOURCE_DIR}/src/integrationtesting.cpp
+	${CMAKE_CURRENT_SOURCE_DIR}/src/scenarioexecution.cpp
+)
+# Link project executable to util libraries
+target_link_libraries(${INTEGRATION_TESTING_TARGET} 
+	${ATOS_COMMON_LIBRARY}
+	${ATOS_COMMON_LIBRARY}
```
SebastianSvanland: Why twice?
```suggestion
	${ATOS_COMMON_LIBRARY}
```

modules/IntegrationTesting/inc/integrationtesting.hpp
```diff
@@ -0,0 +1,36 @@
...
+		~IntegrationTesting();
+
+		static inline std::string const testName = "scenario_execution";
+		virtual void runIntegrationTest() = 0;
+
+	protected:
+	  static inline std::string const initTopic = "/atos/init";
+	  static inline std::string const connectTopic = "/atos/connect";
+	  static inline std::string const armTopic = "/atos/arm";
+	  static inline std::string const startTopic = "/atos/start";
```
SebastianSvanland: Why are these not fetched from ROS channels?

modules/IntegrationTesting/src/integrationtesting.cpp
```diff
@@ -0,0 +1,77 @@
...
+ * 
+ * 
+ * @param moduleName 
+ */
+IntegrationTesting::IntegrationTesting(const std::string& moduleName) : Module(moduleName) {
+	initPub = this->create_publisher<std_msgs::msg::Empty>(initTopic, 10);
+	connectPub = this->create_publisher<std_msgs::msg::Empty>(connectTopic, 10);
+	armPub = this->create_publisher<std_msgs::msg::Empty>(armTopic, 10);
+	startPub = this->create_publisher<std_msgs::msg::Empty>(startTopic, 10);
+	getObjectControlStateClient = this->create_client<atos_interfaces::srv::GetObjectControlState>("/atos/get_object_control_state");
```
SebastianSvanland: This should also be taken from module.hpp where all the services are listed.

modules/IntegrationTesting/src/integrationtestinghandler.cpp
```diff
@@ -0,0 +1,55 @@
...
+void IntegrationTestingHandler::executeIntegrationTests() {
+	std::this_thread::sleep_for(std::chrono::seconds(3)); // sleep thread to allow other nodes to start
+
+	for (auto const& [testName, testEnabled] : integrationTests) {
+		if (testEnabled) {
+			auto executor = std::make_shared<rclcpp::executors::SingleThreadedExecutor>();
+			auto integrationTestFactory = std::make_shared<IntegrationTestingFactory>();
+			auto integrationTest = integrationTestFactory->createIntegrationTestExecution(testName);
+			executor->add_node(integrationTest);
+			executor->spin_once();
```
SebastianSvanland, line 51: Is one spin enough to finish the integration tests? I'm not sure I follow what's going on here (maybe I don't understand executors entirely)

modules/IntegrationTesting/src/scenarioexecution.cpp
```diff
@@ -0,0 +1,154 @@
...
+using namespace std::chrono_literals;
+
+/**
+ * @brief An integration test for scenario execution. This test will run a scenario execution for one object,
+ * and check that all states are correct and that it follows the trajectory.
+ * 
+ */
+ScenarioExecution::ScenarioExecution() : IntegrationTesting(testName) {
+	RCLCPP_INFO(get_logger(), "Started integration test: %s", testName.c_str());
+	getObjectTrajectoryClient = this->create_client<atos_interfaces::srv::GetObjectTrajectory>("/atos/get_object_trajectory");
```
SebastianSvanland: Again, should be fetched from module.hpp where all services are listed.

modules/IntegrationTesting/src/integrationtesting.cpp
```diff
@@ -0,0 +1,77 @@
...
+}
+
+
+/**
+ * @brief Check if the current state of the object is the same as the expected state.
+ * This function also saves the result for printing later.
+ * 
+ * @param command The command sent in order to change state, e.g. "/atos/init".
+ */
+void IntegrationTesting::checkState(const std::string& command) {
```
SebastianSvanland, line 53: Isn't this method receiving topic names rather than commands? Argument feels a little bit misleading.

modules/IntegrationTesting/src/scenarioexecution.cpp
```diff
@@ -0,0 +1,154 @@
...
+ */
+void ScenarioExecution::checkTrajectory() {
+	auto trajectory = getTrajectoryPoints();
+	std::pair<double, double> lastPoint;
+	auto isObjectMoving = true;
+	while(isObjectMoving) {
+		atos_interfaces::msg::Monitor monr;
+		monitorSub = this->create_subscription<atos_interfaces::msg::Monitor>("/atos/object_1/object_monitor", 10, std::bind(&ScenarioExecution::placeholderCallback, this, std::placeholders::_1));
+		rclcpp::wait_for_message(monr, monitorSub, this->get_node_base_interface()->get_context(), 50ms);
+		if (monr.velocity.twist.linear.x == 0 && monr.velocity.twist.linear.y == 0) {
```
SebastianSvanland: You should probably add some leeway to the check here since objects might not always report exactly 0 velocity.
```suggestion
                speedThreshold = 0.1
		if (abs(monr.velocity.twist.linear.x) < speedThreshold && abs(monr.velocity.twist.linear.y) < speedThreshold) {
```

modules/IntegrationTesting/src/scenarioexecution.cpp
```diff
@@ -0,0 +1,154 @@
...
+	RCLCPP_INFO(get_logger(), ss.str().c_str());
+}
+
+
+/**
+ * @brief This is a placeholder callback function. It is needed in order to create a topic subscriber.
+ * 
+ * @param msg Message.
+ */
+void ScenarioExecution::placeholderCallback(const atos_interfaces::msg::Monitor::SharedPtr msg) {}
```
SebastianSvanland, line 155: Why placeholderCallback and not monrCallback?

SebastianSvanland (commented):
Nice work, some minor remarks

victorjarlow (approved):
Aside from Sebastian's comments I don't have anything to add. 
I still feel like we are doing something redundant when we are producing our own error report, like gtest or pytest. Maybe we have duplicated some work here that other open source alternatives solve for us.

Next time we do some work on testing we could investigate other testing frameworks as well, to ensure that we have a scalable way of creating tests.

modules/IntegrationTesting/src/integrationtestinghandler.cpp
samuelthoren: Yes, it should be one spin otherwise if we have multiple integration tests they won't run.

modules/IntegrationTesting/src/scenarioexecution.cpp
samuelthoren: This is just a function needed for `monitorSub` in `rclcpp::wait_for_message`, this function doesn't do anything but is needed anyway. I named it `placeholderCallback` so it's clear it's just a placeholder.

PR #591, opened 2023-09-28
Fixed bug where install_atos.sh wouldn't source correctly at the end,…

SebastianSvanland:
… resulting in other errors.


victorjarlow (approved):

install_atos.sh
```diff
@@ -307,12 +307,12 @@ case "$SHELL" in
     */bash)
         add_source_line_if_needed ~/.bashrc "bash" "${atos_setup_script}"
         add_source_line_if_needed ~/.bashrc "bash" "${ros2_setup_script}"
-        source ~/.bashrc
+        source $HOME/.bashrc
```
victorjarlow:
```suggestion
        source $HOME/.bashrc 
```

PR #596, opened 2023-10-04
Fix proj of trajs

Robert108:
This PR includes:
* Add projection transformation in esmini
* Proj lib to support transforms
* Removes setting lat, long etc from params.yaml (Which was unsupported anyway)
* Read lat long from ODR file instead. 
* Switch getOrigin service to be inside esmini instead.


modules/ObjectControl/src/objectcontrol.cpp
```diff
@@ -351,6 +351,7 @@ void ObjectControl::loadScenario() {
 	// TODO query scenario participants from scenario manager
 	auto request = std::make_shared<atos_interfaces::srv::GetObjectIds::Request>();
 	auto promise = idClient->async_send_request(request, idsCallback);
+	promise.wait_for(std::chrono::seconds(10));
```
samuelthoren: I have just started testing, but when I press `init` it takes 10 seconds for ATOS to get to the initialised state and I think that this is why. When you do this I think you block the thread for 10 seconds, which isn't what we want to do.

samuelthoren (changes requested):

cmake/modules/FindesminiRMLib.cmake
samuelthoren, line 1: Is it possible to specify which version of this library to install here? So that we don't get the problem we discussed on Teams.

docs/Installation/installation.md
```diff
@@ -99,10 +99,10 @@ cd ATOS
 git submodule update --init --recursive
 ```
 
-Install the dependencies:
+Install the dependencies (Stand in the repo directory):
```
samuelthoren: Specify this, in the root of the directory or anywhere?

docs/Usage/Modules/EsminiAdapter.md
```diff
@@ -3,6 +3,7 @@ Module for interfacing with the [esmini](https://github.com/esmini/esmini) Scena
 ## About the module
 This module has the following responsibilities:
     * Load an OpenSCENARIO-file and extract static trajectories from it.
+    * From the associated OpenDrive file, extract the test origin and make available to other modules.
```
samuelthoren:
```suggestion
    * From the associated OpenDRIVE file, extract the test origin and make available to other modules.
```

docs/Usage/Modules/EsminiAdapter.md
```diff
@@ -17,4 +18,14 @@ Load an OpenSCENARIO-file called `MyScenario.xosc`.
 
 - `open_scenario_file: "MyScenario.xosc"`
 
+## Test origin
 
+The test origin is extracted from the OpenDrive file of the scenario. To change the test origin to a different location, change the "geoReference" tag in the OpenDrive file header.
```
samuelthoren:
```suggestion
The test origin is extracted from the OpenDRIVE file of the scenario. To change the test origin to a different location, change the "geoReference" tag in the OpenDRIVE file header. 
```

modules/EsminiAdapter/PROJOptions.cmake
samuelthoren, line 1: Is it necessary for this to be its own file? Can't it be included directly in the CMakeLists?

modules/EsminiAdapter/src/CRSTransformation.cpp
```diff
@@ -0,0 +1,73 @@
...
+	// Put transformed PJ_POINTS back into TrajPoints
+	for (int i = 0; i < arraySize; i++){
+		trajPoints[i].setXCoord(in[i].xyz.x);
+		trajPoints[i].setYCoord(in[i].xyz.y);
+		trajPoints[i].setZCoord(in[i].xyz.z);
+	}
+}
+
+/**
+ * @brief Get the origin lat, lon, height in the given datum from a proh string
```
samuelthoren:
```suggestion
 * @brief Get the origin lat, lon, height in the given datum from a proj string
```

modules/EsminiAdapter/src/CRSTransformation.cpp
samuelthoren, line 1: This is very good, however I do think that this should be moved to the `common` directory since this might be used in other modules and not only `EsminiAdapter`. There is already a `coordinateutils` under `common -> utils`, maybe we can merge these?

samuelthoren (commented):
Very nice! It's good to not having to specify the origin manually, and nice documentation.

I think we have to solve the service call in some way so that we don't block everything else for 10 seconds. I have also added some other minor feedback.

modules/EsminiAdapter/PROJOptions.cmake
Robert108: It is actually not needed at all since we run the system installed version of proj.

modules/EsminiAdapter/src/CRSTransformation.cpp
Robert108: I moved this to util as suggested.

samuelthoren (approved):
Approved!

docs/Usage/Modules/EsminiAdapter.md
```diff
@@ -17,4 +18,14 @@ Load an OpenSCENARIO-file called `MyScenario.xosc`.
...
 
+## Test origin
 
+The test origin is extracted from the OpenDRIVE file of the scenario. To change the test origin to a different location, change the "geoReference" tag in the OpenDrive file header. 
+
+*Example*
+```xml
+<OpenDRIVE>
+  <header ... >
+    <geoReference><![CDATA[+proj=tmerc +lat_0=57.77752905043481 +lon_0=12.7814573051027 +k=1 +x_0=0 +y_0=0 +datum=WGS84 +units=m +geoidgrids=egm96_15.gtx +vunits=m +no_defs ]]></geoReference>
```
victorjarlow, line 29: An optional geoidgrids file location could be specified in params.yaml for example. When I try to run dev right now it will not work until i place a egm96_15.gtx file within my workspace (My scenario has a projstring that requires a geoidgrid file).

victorjarlow: @Robert108

PR #597, opened 2023-10-04
Fix back to start

SebastianSvanland:
Added the BackToStart module as a ROS node in ATOS. Includes functionality for resetting all test objects to their starting position and reload scenario trajectory. 
Need to move path publishers from BackToStart module to objectControl since the trajectory will only update in the gui when calculating the back-to-start trajectory and not when reloading the original trajectory.


modules/BackToStart/CMakeLists.txt
```diff
@@ -9,32 +9,40 @@ set(CMAKE_C_STANDARD 11)
...
-set(ISO_22133_LIBRARY MaestroISO22133)
-set(POSITIONING_LIBRARY MaestroPositioning)
-set(TIME_LIBRARY ATOSTime)
-set(ATOS_COMMON_LIBRARY ATOSCommon)
-set(LOGGING_LIBRARY MaestroLogging)
-set(MESSAGE_BUS_LIBRARY MaestroMQ)
+# set(COREUTILS_LIBRARY ATOSCoreUtil)
+# set(ISO_22133_LIBRARY MaestroISO22133)
+# set(POSITIONING_LIBRARY MaestroPositioning)
+# set(TIME_LIBRARY ATOSTime)
```
Robert108:
```suggestion
```

modules/BackToStart/CMakeLists.txt
```diff
@@ -9,32 +9,40 @@ set(CMAKE_C_STANDARD 11)
...
 set(ATOS_COMMON_LIBRARY ATOSCommon)
+# set(LOGGING_LIBRARY MaestroLogging)
+# set(MESSAGE_BUS_LIBRARY MaestroMQ)
```
Robert108:
```suggestion
```

modules/BackToStart/CMakeLists.txt
```diff
@@ -9,32 +9,40 @@ set(CMAKE_C_STANDARD 11)
...
-	${TIME_LIBRARY}
-	${COREUTILS_LIBRARY}
-	${LOGGING_LIBRARY}
-        ${FILESYSTEM_LIBRARY}
-        ${ATOS_COMMON_LIBRARY}
+target_link_libraries(${BACK_TO_START}
+	# ${TIME_LIBRARY}
+	# ${COREUTILS_LIBRARY}
+	# ${LOGGING_LIBRARY}
+    # ${FILESYSTEM_LIBRARY}
```
Robert108:
```suggestion
```

modules/BackToStart/inc/backtostart.hpp
```diff
@@ -0,0 +1,32 @@
...
+
+private:
+	static inline std::string const moduleName = "back_to_start";
+
+	rclcpp::Service<atos_interfaces::srv::GetObjectReturnTrajectory>::SharedPtr getObjectReturnTrajectoryService; //!< Service to request object return trajectory
+
+	void onExitMessage(const ROSChannels::ResetTest::message_type::SharedPtr) override;
+    void onReturnTrajectoryRequest(const std::shared_ptr<atos_interfaces::srv::GetObjectReturnTrajectory::Request>,
+                            std::shared_ptr<atos_interfaces::srv::GetObjectReturnTrajectory::Response>);
+    // atos_interfaces::msg::CartesianTrajectory cutOffTraj(const CartesianTrajectory& traj, const CartesianPosition& point);
```
Robert108:
```suggestion
```

modules/BackToStart/src/backtostart.cpp
```diff
@@ -0,0 +1,63 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ */
+#include "backtostart.hpp"
+using std::placeholders::_1, std::placeholders::_2;
+
+BackToStart::BackToStart() : Module(BackToStart::moduleName),
```
Robert108: Since this is a a new major module when should create tests for it so that we fulfill our own guidelines :) 
You can take a look at https://github.com/RI-SE/ATOS/pull/599 for inspiration.

modules/BackToStart/src/backtostart.cpp
```diff
@@ -0,0 +1,63 @@
...
+    rclcpp::shutdown();
+}
+
+
+void BackToStart::onReturnTrajectoryRequest(const std::shared_ptr<atos_interfaces::srv::GetObjectReturnTrajectory::Request> request, 
+                                            std::shared_ptr<atos_interfaces::srv::GetObjectReturnTrajectory::Response> response) {
+    ATOS::Trajectory currentTraj(get_logger());
+    currentTraj.initializeFromCartesianTrajectory(request->trajectory);
+    uint32_t id = request->id;
+    // object_position = request->position;
```
Robert108:
```suggestion
```

modules/BackToStart/src/backtostart.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+    // auto trajShortened = cutOffTraj(request->trajectory, object_position);
```
Robert108:
```suggestion
```

modules/BackToStart/src/backtostart.cpp
```diff
@@ -0,0 +1,63 @@
...
+using namespace atos_interfaces::msg;
+CartesianTrajectory BackToStart::cutOffTraj(const CartesianTrajectory& traj, const CartesianPosition& point) {
+    CartesianTrajectory trajShortened;
+    auto it = std::find_if(traj.begin(), traj.end(), [point](const CartesianPoint& p) {
+        return p.getTime() > point.getTime();
+    });
+    trajShortened.insert(trajShortened.end(), it, traj.end());
+    return trajShortened;
+}
+*/
```
Robert108:
```suggestion

```

Robert108: Double check (hint. Maybe with a test) or remove

common/module.hpp
```diff
@@ -72,6 +73,8 @@ class Module : public rclcpp::Node {
 	virtual void onAbortMessage(const ROSChannels::Abort::message_type::SharedPtr){};
 	virtual void onReplayMessage(const ROSChannels::Replay::message_type::SharedPtr){};
 	virtual void onExitMessage(const ROSChannels::Exit::message_type::SharedPtr);
+	virtual void onResetTestMessage(const ROSChannels::ResetTest::message_type::SharedPtr){};
+	virtual void onReloadSettingsMessage(const ROSChannels::ReloadSettings::message_type::SharedPtr){};
```
samuelthoren: I don't think `Reload Settings` is a clear enough name, because we only reset the trajectory, or am I wrong? Maybe it could be named something like `Reset Trajectory` making it more clear what it does.

gui/controlpanel/public/javascripts/client.js
```diff
@@ -68,15 +68,15 @@ function updateATOSStatusIcon(status, new_text) {
...
-function sendResetTest(){
-    sendCommand("send_reset_test", ws);
-}
 function sendInit(){
     sendCommand("send_init", ws);
 }
 function sendConnect(){
     sendCommand("send_connect", ws);
 }
+function sendDisconnect(){
```
samuelthoren, line 77: Might be out of scope for this PR, but pressing this button crashes object control. We should maybe not make it possible to press this button until this issue is fixed.

gui/controlpanel/views/index.pug
```diff
@@ -18,11 +18,14 @@ body(style="padding-bottom:10rem;")
     .d-grid(class="gap-1 d-sm-block")
       +button("info", "sm", onClick='sendInit()') Init
       +button("info", "sm", onClick='sendConnect()') Connect
-      +button("secondary", "sm", onClick='sendResetTest()') Disconnect
+      +button("secondary", "sm", onClick='sendDisconnect()') Disconnect
       +button("warning", "sm", onClick='sendArm()') Arm
       +button("warning", "sm", onClick='sendDisarm()') Disarm
       +button("success", "sm", onClick='sendStart()') Start
       +button("secondary", "sm", onClick='sendAllClear()') Clear Abort
+    .d-grid(class="gap-1 d-sm-block")
```
samuelthoren, line 26: This might be hard to do, but is it possible to only show these buttons only when launching the experimental nodes? Because these buttons don't do anything when launching ATOS "normally", which could lead to confusion.

launch/launch_experimental.py
```diff
@@ -33,6 +33,13 @@ def get_experimental_nodes():
             executable='osi_adapter',
             name='osi_adapter',
             parameters=[files["params"]]
+        ),
+        Node(
+            package='atos',
+            namespace='atos',
+            executable='back_to_start',
+            name='back_to_start',
+            parameters=[files["params"]]
```
samuelthoren:
```suggestion
```
This row isn't necessary unless you want to read parameters from `params.yaml`.

modules/BackToStart/CMakeLists.txt
```diff
@@ -9,32 +9,40 @@ set(CMAKE_C_STANDARD 11)
...
 set(BACK_TO_START ${PROJECT_NAME})
 
-set(COREUTILS_LIBRARY ATOSCoreUtil)
-set(ISO_22133_LIBRARY MaestroISO22133)
-set(POSITIONING_LIBRARY MaestroPositioning)
-set(TIME_LIBRARY ATOSTime)
-set(ATOS_COMMON_LIBRARY ATOSCommon)
-set(LOGGING_LIBRARY MaestroLogging)
-set(MESSAGE_BUS_LIBRARY MaestroMQ)
+# set(COREUTILS_LIBRARY ATOSCoreUtil)
```
samuelthoren: Should the commented rows be here, otherwise remove them.

modules/BackToStart/inc/backtostart.hpp
samuelthoren:
```suggestion
```

modules/BackToStart/src/backtostart.cpp
```diff
@@ -0,0 +1,63 @@
...
+using std::placeholders::_1, std::placeholders::_2;
+
+BackToStart::BackToStart() : Module(BackToStart::moduleName), 
+	exitSub(*this, std::bind(&BackToStart::onExitMessage, this, _1))
+{
+	getObjectReturnTrajectoryService = create_service<atos_interfaces::srv::GetObjectReturnTrajectory>(ServiceNames::getObjectReturnTrajectory,
+		std::bind(&BackToStart::onReturnTrajectoryRequest, this, _1, _2));
+}
+void BackToStart::onExitMessage(const ROSChannels::Exit::message_type::SharedPtr)
+{
```
samuelthoren:
```suggestion
void BackToStart::onExitMessage(const ROSChannels::Exit::message_type::SharedPtr) {
```

modules/BackToStart/src/backtostart.cpp
samuelthoren: Remove the code you've commented out, also the one at the bottom of the file if you shouldn't keep it.

modules/BackToStart/src/main.cpp
```diff
@@ -3,213 +3,12 @@
...
-			LogMessage(LOG_LEVEL_INFO,"Received COMM_BACKTOSTART_CALL command");
-			backToStart();
-			break;
-		default:
-			break;
-		}
-	}
+int main(int argc, char** argv){
+	rclcpp::init(argc, argv);
+	static std::shared_ptr<BackToStart> bts_node = std::make_shared<BackToStart>();
```
samuelthoren: Does this need to be static?

samuelthoren: Use camelCase for variable names!

modules/ObjectControl/inc/objectcontrol.hpp
```diff
@@ -314,6 +335,18 @@ class ObjectControl : public Module
 	void disarmObjects();
 	//! \brief
 	void startScenario();
+	//! \brief TODO
```
samuelthoren: Don't forget TODO!

modules/ObjectControl/src/objectcontrol.cpp
```diff
@@ -169,6 +172,20 @@ void ObjectControl::onStopMessage(const Stop::message_type::SharedPtr){
 	this->tryHandleMessage(f_try,f_catch, Stop::topicName, get_logger());	
 }
 
+void ObjectControl::onResetTestMessage(const ResetTest::message_type::SharedPtr) {
+	COMMAND cmd = COMM_BACKTOSTART_CALL;
```
samuelthoren, line 212: I noticed that it is the same command here and at `onReloadSettingsMessage`, is this correct?

modules/ObjectControl/src/objectcontrol.cpp
```diff
@@ -690,6 +714,91 @@ void ObjectControl::startScenario() {
...
+	RCLCPP_INFO(get_logger(), "Reloading scenario trajectories");
+	for (auto& id : getVehicleIDs()) {
+		this->setObjectTrajectory(id);
+	}
+}
+
+void ObjectControl::updateTrajectoryGUI(uint32_t id){
+	ATOS::Trajectory traj = objects.at(id)->getTrajectory();
+	this->pathPublishers.emplace(id, ROSChannels::Path::Pub(*this, id));
+	this->pathPublishers.at(id).publish(traj.toPath());
```
samuelthoren, line 813: This doesn't seem to work correctly, when I press `Abort` -> `Clear Abort` -> `Reset Test`, the back to start trajectory shows up for only a short time in the 3D-view, but then the original trajectory is shown again. For the map view however, it's working correctly.

modules/ObjectControl/src/objectcontrol.cpp
```diff
@@ -690,6 +714,91 @@ void ObjectControl::startScenario() {
 	}
 }
```
samuelthoren, line 765: Add documentation to all functions

samuelthoren (changes requested):
Good work! It's nice to have this functionality back. I have added some comments, the biggest issue is that the back to start trajectory doesn't seem to show up correctly in the 3D-view. It shows up for about a second, and then I get the original trajectory again.

modules/EsminiAdapter/src/esminiadapter.cpp
```diff
@@ -539,8 +539,8 @@ void EsminiAdapter::InitializeEsmini()
 
 	me->extractTrajectories(0.1, me->idToTraj);
 
-	RCLCPP_INFO(me->get_logger(), "Extracted %d trajectories", me->idToTraj.size());
-	RCLCPP_INFO(me->get_logger(), "Number of objects with triggered start: %d", me->delayedStartIds.size());
+	RCLCPP_INFO(me->get_logger(), "Extracted %ld trajectories", me->idToTraj.size());
```
Robert108, line 687: Nice!

common/module.hpp
Robert108: Agreed

gui/controlpanel/views/index.pug
Robert108: Yes, We should add in any case a condition to the buttons that they are not clickable in curtain states. Maybe the BTS buttons can listen for msgs being published that are specific to the BTS module.

Robert108 (changes requested):
Great to have BTS available again! Remember to add Doxygen comments and ReadTheDocs documentation.

common/module.hpp
SebastianSvanland: The idea here is that the reload settings is supposed to reload everything that is needed to run the test again, it just so happens that it is only the trajectory that has been changed during the reset but could potentially be more so I wanted to keep it open. However, I do agree that these names (reset test included) are difficult for a user to understand and that maybe it should be done automatically when the object has returned to its starting position so you only need to click on one button for everything (reset test in this case which would reload settings as well). Though that would require a lot more functionality with triggers on positions etc which isn't really implemented in ATOS yet. 

Let me know if you still feel like it needs a name change

gui/controlpanel/public/javascripts/client.js
SebastianSvanland: Hmm, I don't experience this. Could it be in combination with something else?

gui/controlpanel/views/index.pug
SebastianSvanland: This is the case for many buttons if pressed in the incorrect order. I would suggest a new task to upgrade the GUI with this kind of functionality

modules/BackToStart/CMakeLists.txt
SebastianSvanland: Oops thought I had removed them when I noticed everything worked

modules/BackToStart/src/main.cpp
SebastianSvanland: I think the static was remainder from copying some code, doesn't need to be there so I removed it. Also changed to camelCase

modules/ObjectControl/src/objectcontrol.cpp
SebastianSvanland: They are both related to the reset functionality but I changed it to `COMM_OSEM`, should hopefully be more accurate?

modules/ObjectControl/src/objectcontrol.cpp
SebastianSvanland: This functionality actually behaves as it should. However, if you run the experimental launch file with trajectorylet_streamer it will bug out. I'm guessing that the trajectorylet_streamer continues to publish trajectories on the topic.

modules/ObjectControl/src/objectcontrol.cpp
SebastianSvanland: I added some documentation in the hpp-file, do you think it's enough? I noticed there is no documentation for the rest of the code.

SebastianSvanland:
Can't merge until #599 is merged.

SebastianSvanland:
I'm not sure what else to write about the module in the readthedocs, the module itself is very simple. Any suggestions @Robert108 or is it enough?

modules/BackToStart/tests/test_backtostart.cpp
```diff
@@ -0,0 +1,124 @@
...
+    RCLCPP_INFO(rclcpp::get_logger("rclcpp"), "service not available, waiting again...");
+  }
+
+  // Act
+  returnTrajectoryClient->async_send_request(returnTrajectoryRequest, returnTrajResponseCallback);
+  // wait for service to respond and callback to execute
+  test_rclcpp::wait_for_future(executor, service_called_future, fail_after_timeout);
+
+  // Assert
+  ASSERT_EQ(receivedResponse, true);
```
Robert108, line 113: Could you assert that the last point in the back to start traj is actually near the start point in the original traj?

docs/Usage/Modules/BackToStart.md
```diff
@@ -1,4 +1,5 @@
 # BackToStart
+A module for generating an object's return trajectory.
 
 ## About the module
-This module is currently in development and documentation for this module will be released later.
\ No newline at end of file
+This module provides a service to reverse the test object's trajectory and generate a Williamson turn at the start and end to let the object return to its starting point. Execute the new trajectories similar to a normal test with `Arm` and `Start`. It is up to the operator to determine if the new trajectories are possible to execute or not. The service is only allowed in `Disarmed` state and calling `Reset test` in the GUI does not result in a state change.
```
Robert108: "The service is only allowed in `Disarmed` state and calling `Reset test` in the GUI does not result in a state change."
 Sorry i don't understand this. A step by step guide on how the function is used after a test would be awesome!

Robert108 (changes requested):

docs/Usage/Modules/BackToStart.md
SebastianSvanland: I included a step by step guide as well as two images showing what happens when pressing `Reset Test`, hopefully it's more clear now.

common/module.hpp
samuelthoren: Yes, that's a good point. We can maybe keep the name as it is for now.

gui/controlpanel/public/javascripts/client.js
samuelthoren: It happen if I press `Init` -> `Connect` -> `Disconnect`, and I'm using one object. Maybe we can take a look at this together, because it seems to happen everytime for me.

gui/controlpanel/views/index.pug
samuelthoren: I agree, we should make a ticket for this. Maybe the buttons that can't be pressed should be greyed out or something like that.

modules/ObjectControl/src/objectcontrol.cpp
samuelthoren: I see, I think that `TrajectoryletStreamer` will continue publishing the original trajectory. I can add this as a bug, because `TrajectoryletStreamer` should also publish the back to start trajectory.

modules/ObjectControl/src/objectcontrol.cpp
samuelthoren: A lot of the old code doesn't have much documentation, but for the newer code we have documented the code in the `.cpp`-files. See for instance `OSIAdapter` or `EsminiAdapter`.

SebastianSvanland: Added some more documentation, the functions are pretty straight forward in my opinion.

samuelthoren: I meant Doxygen comments for the function descriptions. Comments inside the functions are not really necessary for the most parts.

docs/Usage/Modules/BackToStart.md
```diff
@@ -1,4 +1,20 @@
 # BackToStart
+A module for generating an object's return trajectory.
+
+WARNING: It is up to the operator to determine if the new trajectories are possible to execute or not since the generated trajectories doesn't take other objects' trajectories into consideration.
 
 ## About the module
-This module is currently in development and documentation for this module will be released later.
\ No newline at end of file
+This module provides a service to reverse the test object's trajectory and generate a Williamson turn at the start and end to let the object return to its starting point after a test has been executed.
```
victorjarlow:
```suggestion
This module provides a service that enables resetting test objects to their starting position after a completed test.
The service calculates Williamson turns on both ends of the original trajectory, which if followed by the test object will lead it back to the start of the original trajectory, with a correct heading.
```

common/module.hpp
```diff
@@ -72,6 +73,8 @@ class Module : public rclcpp::Node {
 	virtual void onAbortMessage(const ROSChannels::Abort::message_type::SharedPtr){};
 	virtual void onReplayMessage(const ROSChannels::Replay::message_type::SharedPtr){};
 	virtual void onExitMessage(const ROSChannels::Exit::message_type::SharedPtr);
+	virtual void onResetTestMessage(const ROSChannels::ResetTest::message_type::SharedPtr){};
```
victorjarlow: Could maybe be called ResetTestObjectsMessage

common/module.hpp
victorjarlow: ReloadObjectSettingsMessage

modules/ObjectControl/inc/objectcontrol.hpp
```diff
@@ -70,6 +77,7 @@ namespace AbsoluteKinematics {
 	class Disarming;
 	class Done;
 	class RemoteControlled;
+	class Resetting;
```
victorjarlow: Why is a new Resetting state needed when there is a isResetting member variable?

common/module.hpp
victorjarlow: Remember to change in all places if switching to this name

modules/ObjectControl/inc/objectcontrol.hpp
```diff
@@ -314,6 +335,18 @@ class ObjectControl : public Module
...
+	void resetTest();
+	//! \brief Reloads the scenario trajectories for each object.
+	void reloadScenarioTrajectories();
+	//! \brief Updates the paths in the GUI to reflect the new trajectories.
+	void updateTrajectoryGUI(uint32_t id);
+	//! \brief Callback for the trajectory request. Sends the new trajectory to the object.
+	void trajectoryCallback(const rclcpp::Client<atos_interfaces::srv::GetObjectTrajectory>::SharedFuture future);
+	//! \brief Callback for the return trajectory request. Sends the new trajectory to the object.
+	void returnTrajectoryCallback(const rclcpp::Client<atos_interfaces::srv::GetObjectReturnTrajectory>::SharedFuture future);
+	//! \brief Requests a new trajectory to the object depending on the current state (resetting).
```
victorjarlow: Requests a new trajectory? The below method is called setObjectTrajectory. Either of comment/method name does not describe what actually happens in the method.

modules/ObjectControl/src/objectcontrol.cpp
```diff
@@ -690,6 +714,107 @@ void ObjectControl::startScenario() {
...
+	}
+	RCLCPP_INFO(get_logger(), "Reloading scenario trajectories");
+	// Request the scenario trajectory for each object
+	for (auto& id : getVehicleIDs()) {
+		this->setObjectTrajectory(id);
+	}
+	this->isResetting = false;
+}
+
+void ObjectControl::updateTrajectoryGUI(uint32_t id){
```
victorjarlow: Could be called e.g. republishTrajectories. The GUI is just one of many potential "listeners" on the topic.

modules/ObjectControl/src/objectcontrol.cpp
```diff
@@ -690,6 +714,107 @@ void ObjectControl::startScenario() {
...
+	// Update the GUI with the new trajectory in local coordinates
+	ATOS::Trajectory traj = objects.at(id)->getTrajectory();
+	this->pathPublishers.emplace(id, ROSChannels::Path::Pub(*this, id));
+	this->pathPublishers.at(id).publish(traj.toPath());
+
+	// Update the GUI with the new trajectory in global coordinates
+	GeographicPositionType origin = objects.at(id)->getOrigin();
+	std::array<double,3> llh_0 = {origin.latitude_deg, origin.longitude_deg, origin.altitude_m};
+	this->gnssPathPublishers.emplace(id, ROSChannels::GNSSPath::Pub(*this, id));
+	this->gnssPathPublishers.at(id).publish(traj.toGeoJSON(llh_0));
```
victorjarlow, line 819: We need to implement a simple "republisher" node that takes CartesianTrajectories as input and republishes them as GnssPaths... But It is not part of this PR.

victorjarlow (approved):
Good job, I left some comments mainly relating to clarifying names. The next step in this work is to incorporate a resetting strategy (i.e. operator determines if all should reset at the same time, or object by object, and so on..)

Robert108 (commented):
Looks good, I made some changes to the sample module unit tests and pushed matching changes to this PR.

modules/ObjectControl/inc/objectcontrol.hpp
SebastianSvanland: Remains from another approach that I forgot to remove. Fixed now

modules/ObjectControl/src/objectcontrol.cpp
SebastianSvanland: Renamed in new commit

modules/ObjectControl/inc/objectcontrol.hpp
SebastianSvanland: Updated the documentation, let me know if you're satisfied

Robert108 (approved):

samuelthoren (approved):
Good job on the PR! It doesn't crash anymore when I press disconnect, however in Foxglove it still says it is connected when the object has disconnected when I press the button. Please fix this before you merge.

SebastianSvanland:
> Good job on the PR! It doesn't crash anymore when I press disconnect, however in Foxglove it still says it is connected when the object has disconnected when I press the button. Please fix this before you merge.

Once again, I do not experience the same issues as you do. Maybe we can look at it together

PR #600, opened 2023-10-10
Fix update install scripts

Robert108:
This PR uses the install scripts inside the Dockerfile which both deletes repeated code and creates a quick way of validating the install scripts since they now will be executed every time the docker image is build.

It also splits the install script into two to make use of the Docker image layer caching.

A new user entry point script is created in the root folder called setup_atos.sh


Robert108:
I have added the docker build as a check here to see if it is reasonable to build it every time a commit is made in a PR towards dev.

.github/workflows/build.yaml
```diff
@@ -3,6 +3,8 @@ name: Docker Image CI
 on:
   push:
     branches: [ "dev*" ]
+  pull_request:
+    branches: [ "dev" ]
```
samuelthoren, line 7: Should we add the `master` branch here too?

samuelthoren (approved):
Approved by me, but I think someone with more experience with Docker than me should also approve before merging!

scripts/install_atos.sh
```diff
@@ -0,0 +1,104 @@
...
+################################################
+###### Install Control Panel dependencies ######
+################################################
+
+echo "Installing nvm and Node.js..."
+if [ "$ROS_DISTRO" == "humble" ]; then 
+    NODE_VERSION=16.20.0    # Also specified in the Dockerfile
+elif [ "$ROS_DISTRO" == "foxy" ]; then
+    NODE_VERSION=12
+fi
```
victorjarlow: What should be in the default else case here?

scripts/install_deps.sh
```diff
@@ -0,0 +1,116 @@
...
+    ros-${ROS_DISTRO}-foxglove-msgs \
+    ros-${ROS_DISTRO}-pcl-conversions \
+    ros-${ROS_DISTRO}-rosbridge-suite
+check_command_failed $? "Failed to install ROS2 packages."
+
+if [ "$ROS_DISTRO" == "humble" ]; then
+    sudo apt install -y libpaho-mqtt-dev
+elif [ "$ROS_DISTRO" == "foxy" ]; then
+    sudo apt install -y ros-foxy-paho-mqtt-c
+fi
```
victorjarlow: What happens in case of neither humble nor foxy?

setup_atos.sh
```diff
@@ -0,0 +1,77 @@
...
+
+# Set ROS_DISTRO based on Ubuntu distribution
+case "$(get_ubuntu_codename)" in
+    "focal")
+        export ROS_DISTRO=foxy
+    ;;
+    "jammy")
+        export ROS_DISTRO=humble
+    ;;
+    *)
```
victorjarlow: Is this really correct? I think that you can have ubuntu 20 and humble for example.. Would this case break things further downstream?

victorjarlow: The ROS_DISTRO env variable is set/will be overwritten when sourcing /opt/ros/distro/setup.zsh
If a ros distro is already in use then running this script might overwrite this variable in the active environment of the user. Is there a reason for overwriting actual environment variables such as this one and NVM_DIR, or could temporary "script-local" variables be used (and cleaned up afterwards)?

scripts/install_atos.sh
```diff
@@ -0,0 +1,104 @@
...
+# Check if called without arguments
+if [ $# -eq 0 ]; then
+    echo "Don't call this file directly, use setup_atos.sh instead."
+    echo "Usage: ./install_deps.sh <path to ATOS git repo>"           
+    exit 0
+fi
+
+# Take the first argument as the PATH to the ATOS git repo
+ATOS_REPO_PATH="$1"
+source "${ATOS_REPO_PATH}/scripts/install_functions.sh"
```
victorjarlow, line 13: check_command_failed could be added here to abort the program if an incorrect ATOS_REPO_PATH was input

scripts/install_deps.sh
```diff
@@ -0,0 +1,116 @@
...
+
+# Install OpenSimulationInterface but first check if the library is already installed
+if [ -d "/usr/local/lib/osi3" ]; then
+    echo "OpenSimulationInterface already installed, skipping installation..."
+else
+    echo "Installing OpenSimulationInterface..."
+    git clone --depth 1 --branch v3.4.0 https://github.com/OpenSimulationInterface/open-simulation-interface.git $SOURCE_PATH/open-simulation-interface
+    cd $SOURCE_PATH/open-simulation-interface
+    mkdir -p build && cd build
+    cmake .. && make -j8
```
victorjarlow, line 81: can hardcoded j8 cause problems here? maybe use nproc

victorjarlow (commented):
Good idea to consolidate the installation into one place (setup_atos.sh) instead of having it both in scripts and in dockerfile!

scripts/install_atos.sh
```diff
@@ -0,0 +1,104 @@
...
+echo "Updating submodules to make sure they are up to date..."
+cd $HOME/atos_ws/src/atos
+git submodule update --init --recursive
+cd -
+
+# temporarily cd into the workspace and build with colcon
+echo "Building ATOS..."
+cd $HOME/atos_ws
+source /opt/ros/$ROS_DISTRO/setup.bash
+MAKEFLAGS=-j4 && colcon build --symlink-install
```
victorjarlow: max of j4 and nproc? select number of threads based on output of free -h?

victorjarlow: "MAKEFLAGS=-jX colcon build" can be used to only set the MAKEFLAGS variable for the next command (and not pollute the user env as "MAKEFLAGS=-jX && colcon build" does)

Robert108: I hade problems running this script in the Dockerfile with -jX, X > 4.

setup_atos.sh
Robert108: I took (wrongfully) for granted that that ROS distros where basically tied to Ubuntu distros. Since they are not, I removed the Foxy support since we only have one user running foxy still.

scripts/install_deps.sh
Robert108: Causes issues in the Docker build otherwise. Having high j value then available processors will not cause an issue.

scripts/install_deps.sh
Robert108: See other comment

scripts/install_deps.sh
victorjarlow: Maybe github actions builds could be sped up by setting lower jX values. Could be investigated!

PR #601, opened 2023-10-10
Added handling of GREM messages

samuelthoren:
If test mode is set to online mode, we also receive a GREM message. Handling of this message wasn't implemented which led to strange behaviours when setting test mode to online mode. This PR fixes this.


Robert108 (approved):
lgtm

PR #603, opened 2023-10-12
Fix segmentation fault

samuelthoren:
This PR fixes the segmentation fault that occured in the module `TrajectoryletStreamer` when more than one object were connected. There was also a timing issue in the publishing of chunks, meaning that the publishing of chunks all started at the same time. This shouldn't be the case though since usually the objects don't start at the same time. This was solved by listening to the topic `/atos/start_object` and the publishing of chunks now start at the correct times.


SebastianSvanland (approved):
LGTM

PR #606, opened 2023-10-13
remove progress bar from install script

victorjarlow:
removes the spammed idealTree status when building atos, the status spam can be seen in the attached image:
image


samuelthoren (approved):
Very nice!!

PR #607, opened 2023-10-13
Feature foxglove bridge

victorjarlow:
* Updated docs to point users toward the foxglove studio docker implementation instead of studio.foxglove.dev. 
*  Adopted foxglove_bridge as the default bridge, instead of ros_bridge (see motivation here: https://github.com/foxglove/ros-foxglove-bridge#motivation). 
* Added options to launch_basic such that bridge security and implementation can be controlled.
* Updated install instructions and install script
* Logs end up in ~/.ros/logs/date-xx-xxx/launch.log on my system (I have not set ROS_HOME or ROS_LOG_DIR)

example: ros2 launch atos launch_basic.py insecure:=True foxbridge:=True

If no input arguments are specified, the options are set to insecure=False, foxbridge=True.

**Motivation for this PR:**
* The PR removes the need for accepting https://localhost:3443 certificate
* The need for accepting the https://localhost:8765 certificate still remains.. 
* We can change the standard ros launch command to "ros2 launch atos launch_basic insecure:=True" to highlight that the bridge protocol is insecure and that the user can do extra steps (accept cert at https://localhost:8765) in order to simply use the default launch command, but this requires extra steps.


install_atos.sh
```diff
@@ -135,11 +135,11 @@ sudo apt install -y \
     ros-${ROS_DISTRO}-geographic-msgs \
     ros-${ROS_DISTRO}-foxglove-msgs \
     ros-${ROS_DISTRO}-pcl-conversions \
-    ros-${ROS_DISTRO}-rosbridge-suite
+    ros-${ROS_DISTRO}-rosbridge-suite \
```
victorjarlow:
```suggestion
    ros-${ROS_DISTRO}-rosbridge-suite
```

docker-compose.yml
```diff
@@ -13,12 +13,17 @@ services:
...
-      - "8080:8080" 
       - "8081:8081"
       - "8082:8082"
       - "3000:3000" 
       - "3443:3443"
       - "55555:55555"
       - "443:443"
       - "9090:9090" 
-    command: bash -c "source /root/atos_ws/install/setup.sh ; ros2 launch atos launch_basic.py"
+    command: bash -c "source /root/atos_ws/install/setup.sh ; ros2 launch atos launch_basic.py insecure:=${INSECURE} foxbridge:=${USE_FOXBRIDGE}"
```
Robert108: I added this for those running ATOS in docker

docs/Usage/GUI/foxglove.md
```diff
@@ -1,8 +1,11 @@
 # Foxglove Studio
-If you want to run ATOS with more visualisation you can use [Foxglove Studio](https://github.com/foxglove/studio). To get started, visit [studio.foxglove.dev](https://studio.foxglove.dev).
+If you want to run ATOS with more visualisation you can use [Foxglove Studio](https://github.com/foxglove/studio). 
+Run the following docker command to start the GUI:
+`docker run --rm -p "8080:8080" ghcr.io/foxglove/studio:latest`
+Open a browser and navigate to `http://localhost:8080`
```
Robert108, line 6: We should add some notes about how to run with ssl and how to accept the self signed keys

launch/launch_utils/launch_base.py
```diff
@@ -73,5 +71,72 @@ def get_base_nodes():
...
             name='esmini_adapter',
             parameters=[files["params"]]
+        ),
+        Node(
+            condition=IfCondition(PythonExpression(['not ', foxbridge])),
+            name='rosapi',
+            package='rosapi',
+            executable='rosapi_node'
+        ),
+        Node(
```
Robert108, line 99: A lot of duplication here. Can it be rewritten?

launch/launch_utils/launch_base.py
```diff
@@ -17,42 +20,37 @@ def get_files():
...
 
     return [
+        foxbridge_launch_arg,
+        insecure_launch_arg,
+        ExecuteProcess(
+            name='control_gui',
+            output={'both': 'log'}, #print to log to avoid cluttering the terminal
+            cmd=[[
+                FindExecutable(name='npm'),
+                ' start --prefix ',
```
Robert108, line 60: It would be nice to to see stderr in the terminal to know if node crashes for any reason.

Robert108 (changes requested):
Really nice that we can avoid the problems a lot of people are having with the certificates with this.

launch/launch_utils/launch_base.py
victorjarlow: Using the currently implemented method, the user is made aware of crashes 
MicrosoftTeams-image (13)

samuelthoren (approved):
I agree with Robert's comments, otherwise it looks good to merge!

victorjarlow:
@Robert108  requesting re-review

Robert108 (approved):

PR #609, opened 2023-10-18
Check for if scenario engine should quit

samuelthoren:
This fixes some of the issues where we get a simulation time of 3600.1 seconds. This works for when you set a stop trigger on the time in a scenario. I haven't tried on any other type of stop trigger yet, so I don't know if it works on them too.

### DIFF
--- a/modules/EsminiAdapter/src/esminiadapter.cpp
+++ b/modules/EsminiAdapter/src/esminiadapter.cpp
@@ -564,6 +564,10 @@ void EsminiAdapter::getObjectStates(
 	constexpr double MAX_SCENARIO_TIME = 3600.0;
 	bool stopSimulation = false;
 	while (!stopSimulation) {
+		if (SE_GetQuitFlag() != 0) {
+			break;
+		}
+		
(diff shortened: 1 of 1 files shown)


Robert108 (approved):

victorjarlow:
> This fixes some of the issues where we get a simulation time of 3600.1 seconds. This works for when you set a stop trigger on the time in a scenario. I haven't tried on any other type of stop trigger yet, so I don't know if it works on them too.

I think that we should document somewhere how we expect openScenario files to look in order for us to support them (i.e. needs to have a stop-trigger etc)

samuelthoren:
> > This fixes some of the issues where we get a simulation time of 3600.1 seconds. This works for when you set a stop trigger on the time in a scenario. I haven't tried on any other type of stop trigger yet, so I don't know if it works on them too.
> 
> I think that we should document somewhere how we expect openScenario files to look in order for us to support them (i.e. needs to have a stop-trigger etc)

Yes, that's a good idea. I'll make a ticket about this.

PR #611, opened 2023-11-03
Feature automatic test of scenario

Robert108:
This PR builds upon  #589. It introduces: 

- A docker compose file starting a docker version of the ISOObject https://github.com/RI-SE/isoObject the built docker version of ATOS and running the tests. 
- A modified github build and test job running the integration test on each commit.

Some minor changes are also made. 
- The main docker compose file now runs network_mode: host instead of using port mapping. This changes was made since this will be more user friendly for people unfamiliar with docker networking. The port mappings are commented for those that needs it. 
- Minor update to documentation and bug fixes.


SebastianSvanland (approved):
Looks good

PR #612, opened 2023-11-09
Fix the compiler warnings that occurs when building the modules

samuelthoren:
I have fixed all the compiler warnings that occurs with `-Wall -Wextra` when compiling with GCC or Clang. This PR is only for the warnings that occurs when building the modules, which means some of the files in common are also affected. I have not fixed any warnings that occurs when building the common directory or the iso22133-submodule. I think before we fix all compiler warnings there we need to first remove a lot of old code that is no longer used since there is no meaning in fixing code that isn't used.

Please let me know if any of the changes are implemented wrong or breaks anything.


common/objectconfig.hpp
```diff
@@ -41,7 +41,7 @@ class ObjectConfig : public Loggable {
 	GeographicPositionType getOrigin() const { return origin; }
 	std::string getProjString() const;
 	ATOS::Trajectory getTrajectory() const { return trajectory; }
-	void setTrajectory(const ATOS::Trajectory& newTraj) { trajectory = newTraj; } // TODO danger danger - don't do this
+	void setTrajectory(const ATOS::Trajectory& newTraj) { trajectory = newTraj; }
```
victorjarlow, line 44: Why was the comment removed?

common/sockets/socket.hpp
```diff
@@ -112,7 +112,7 @@ class Socket : public BasicSocket
 public:
 
 	using BasicSocket::BasicSocket;
-	Socket(const Socket& other){}
+	Socket(const Socket& other) : BasicSocket(other) {}
```
victorjarlow: This is not semantically equivalent?

modules/JournalControl/src/journalmodelcollection.cpp
```diff
@@ -226,7 +226,7 @@ int JournalModelCollection::dumpToFile(std::string fileName) {
 
 				if (section.end - section.beg < 0) {
 					RCLCPP_ERROR(get_logger(), "End precedes beginning in file %s: beg @%ld, end @%ld",
-							   file.c_str(), section.beg, section.end);
+							   file.c_str(), (long int)section.beg, (long int)section.end);
```
victorjarlow, line 229: Maybe better to change the type in logging output from %ld to whatever fits section.beg?
Instead of casting the values to fit %ld?

Ignore this comment if irrelevant

modules/MQTTBridge/inc/mqtt.hpp
```diff
@@ -30,7 +30,7 @@ class Message {
 
 	Message() {}
 	Message(const std::string& msg, const std::string& topic)
-		: payload(msg), topic(topic) {}
+		: topic(topic), payload(msg) {}
```
victorjarlow, line 33: Just out of interest, why are the arguments switched?

victorjarlow (approved):

modules/MQTTBridge/inc/mqtt.hpp
samuelthoren: This was a very common warning for many of our initialiser lists. The error we got here is `-Wreorder` meaning that the member variables are not in the same order as they are written in the class declaration. This could lead to unexpected behaviours since variables are initialised in the order they are specified in the class declaration and not in the order of the initialiser list.

modules/JournalControl/src/journalmodelcollection.cpp
samuelthoren: Yes I would also like to do that, but in my understanding what type `section.beg` and `section.end` are depends on the system, so we can't know if it should be `%ld` or any other value. So casting this to a long int should be safe I hope.

common/objectconfig.hpp
samuelthoren: I added a copy assignment operator, but maybe this doesn't fix all the problems? Should we keep the comment?

common/sockets/socket.hpp
samuelthoren: Before it called the default constructor for `BasicSocket`, but was that really the intended purpose? The warning we got here was `base class 'class BasicSocket' should be explicitly initialized in the copy constructor`.

PR #615, opened 2023-12-08
Fix/back to start

SebastianSvanland:
Added turnRadius parameter for back to start module. Makes it easier to change how big the turn should be and now scales the speed depending on the radius.

### DIFF
--- a/common/trajectory.cpp
+++ b/common/trajectory.cpp
@@ -499,8 +499,12 @@ Trajectory Trajectory::createWilliamsonTurn(
 	using namespace std::chrono;
 	using Eigen::MatrixXd;
 
-	constexpr double topSpeed = 10 / 3.6;
-	const int calculatedNoOfPoints = 500;
+	double topSpeed = turnRadius*2 / 3.6; // Set top speed depending on radius.
+	// Limit top speed.
+	double minSpeed = 4.0;
+	double maxSpeed = 12.0;
+	std::clamp(topSpeed, minSpeed, maxSpeed);
+	const int calculatedNoOfPoints = 50;
 	double radius = turnRadius;
 	double headingRad = startPoint.getHeading();
 
--- a/common/trajectory.cpp
+++ b/common/trajectory.cpp
@@ -505,7 +505,7 @@ Trajectory Trajectory::createWilliamsonTurn(
 	// Limit top speed.
 	std::clamp(topSpeed, minSpeed, maxSpeed);
 
-	const int calculatedNoOfPoints = 200;
+	const int calculatedNoOfPoints = 200; // TODO: Make this variable dynamic depending on turnRadius
 	double radius = turnRadius;
 	double headingRad = startPoint.getHeading();
 
--- a/common/trajectory.cpp
+++ b/common/trajectory.cpp
@@ -492,6 +492,8 @@ Trajectory Trajectory::rescaledToVelocity(
 Trajectory Trajectory::createWilliamsonTurn(
 		double turnRadius,
 		double acceleration,
+		double minSpeed,
+		double maxSpeed,
 		TrajectoryPoint startPoint,
 		std::chrono::milliseconds startTime)
 {
@@ -501,10 +503,9 @@ Trajectory Trajectory::createWilliamsonTurn(
 
 	double topSpeed = turnRadius*2 / 3.6; // Set top speed depending on radius.
 	// Limit top speed.
-	double minSpeed = 4.0;
-	double maxSpeed = 12.0;
 	std::clamp(topSpeed, minSpeed, maxSpeed);
-	const int calculatedNoOfPoints = 50;
+
+	const int calculatedNoOfPoints = 200;
 	double radius = turnRadius;
 	double headingRad = startPoint.getHeading();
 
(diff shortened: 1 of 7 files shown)


common/trajectory.cpp
```diff
@@ -499,8 +499,12 @@ Trajectory Trajectory::createWilliamsonTurn(
...
 	using Eigen::MatrixXd;
 
-	constexpr double topSpeed = 10 / 3.6;
-	const int calculatedNoOfPoints = 500;
+	double topSpeed = turnRadius*2 / 3.6; // Set top speed depending on radius.
+	// Limit top speed.
+	double minSpeed = 4.0;
+	double maxSpeed = 12.0;
+	std::clamp(topSpeed, minSpeed, maxSpeed);
+	const int calculatedNoOfPoints = 50;
```
victorjarlow: Could these be parameters in params.yaml as well?

modules/BackToStart/src/backtostart.cpp
```diff
@@ -26,6 +27,17 @@ void BackToStart::onReturnTrajectoryRequest(const std::shared_ptr<atos_interface
...
+
+	// Get the turn radius of the williamson turn.
+	double turnRadius;
+	auto success = this->get_parameter("turnRadius", turnRadius);
+	if (!success) {
+		throw std::runtime_error("Could not read parameter turnRadius. Using 5m as default.");
+	}
+	else {
+        turnRadius = 5.0;
+	}
```
victorjarlow: This logic looks like it is inverted.
If we fail to get the turnRadius param, we throw an exception with the text "using 5m as default" and exit the current function.
If we successfully get the turnRadius parameter, we ignore it, and forcefully set turnRadius to 5.0.

victorjarlow (changes requested):

modules/BackToStart/src/backtostart.cpp
samuelthoren: The default value for turn radius is set when declaring the parameter, meaning this isn't necessary. Or am I misunderstanding what's happening here?

samuelthoren (approved):
I agree with Victor's comments, otherwise it looks good!

modules/ObjectControl/src/objectcontrol.cpp
```diff
@@ -875,6 +875,12 @@ void ObjectControl::returnTrajectoryCallback(const rclcpp::Client<atos_interface
 void ObjectControl::setObjectTrajectory(uint32_t id){
 	// Check if object is in resetting "state". If so, offer return trajectory instead of normal trajectory
 	if (this->isResetting) {
+		using namespace std::chrono_literals;
+		if (!this->returnTrajectoryClient->wait_for_service(1s))
+		{
+			RCLCPP_ERROR(get_logger(), "The return trajectory service is not available. Try again later.");
+			return;
+		}
```
victorjarlow: Can this lead to issues since wait_for_service is blocking?

modules/BackToStart/src/backtostart.cpp
```diff
@@ -10,7 +10,9 @@ BackToStart::BackToStart() : Module(BackToStart::moduleName)
 {
 	getObjectReturnTrajectoryService = create_service<atos_interfaces::srv::GetObjectReturnTrajectory>(ServiceNames::getObjectReturnTrajectory,
 		std::bind(&BackToStart::onReturnTrajectoryRequest, this, _1, _2));
-	this->declare_parameter("turnRadius", 5.0);
+	this->declare_parameter("turn_radius", 5.0);
+	this->declare_parameter("min_speed", 4.0);
+	this->declare_parameter("max_speed", 12.0);
```
victorjarlow, line 15: An idea: Variables would be added e.g. 
default_turn_radius = 5.0;
instead of 5.0 in the declare_parameter method.

common/trajectory.cpp
```diff
@@ -492,15 +492,20 @@ Trajectory Trajectory::rescaledToVelocity(
...
 	using namespace std::chrono;
 	using Eigen::MatrixXd;
 
-	constexpr double topSpeed = 10 / 3.6;
-	const int calculatedNoOfPoints = 500;
+	double topSpeed = turnRadius*2 / 3.6; // Set top speed depending on radius.
+	// Limit top speed.
+	std::clamp(topSpeed, minSpeed, maxSpeed);
+
+	const int calculatedNoOfPoints = 200;
```
victorjarlow: Can this hard-coding lead to problems?

modules/ObjectControl/src/objectcontrol.cpp
SebastianSvanland: Not really since I designed it to only look once if the service is available (if-statement instead of while-statement). If it doesn't find the service after just one look, it will just return.

modules/BackToStart/src/backtostart.cpp
SebastianSvanland: Hmm, it seems a bit redundant since the input itself is defined as default value. I have never seen a ros default parameter value be set with a variable before

common/trajectory.cpp
SebastianSvanland: Not sure, it has always been hardcoded, but maybe the number of points should be proportional to the radius, something like turnRadius*100 might be suitable

SebastianSvanland: Made a quick investigation and it seems like it needs to be a constant integer for the matrices created later on to be correct, which basically means that the variable needs to be hardcoded

SebastianSvanland: @victorjarlow  Made an attempt to implement dynamic variable and matrices but it wasn't really possible, primarily the "eigen::array"s were difficult with the rest of the code. I think if you still want the dynamic variable, we can write it as an issue and handle it separately.

victorjarlow (approved):
Approved!

common/trajectory.cpp
SebastianSvanland:
```suggestion
	const int calculatedNoOfPoints = 200; // TODO: Make this variable dynamic depending on turnRadius
```

PR #620, opened 2023-12-20
Feature/configpanel

SebastianSvanland:
New config panel. Go to localhost:3000/configpanel to try it out. No ATOS modules support this gui so far, but the ROS parameters are requested from each node at least. Needed to do some restructuring in the previous controlpanel as well so this PR works towards multiple tickets.

I've been working on this for a while now so I've gone blind for all the variable/function names. Would appreciate if someone could help me clean it up, right now it's a mess. Also check the descriptions/title of all the parameters, I didn't fully understand what all of them did so I just made something up for now. You can either see this in the gui immediately when entering value for each parameter, or check out [json-schema](https://github.com/RI-SE/ATOS/blob/feature/configpanel/gui/src/public/json/atos-param-schema.json)

There is a known issue where the config gui doesn't send the ros parameters for all nodes. Don't really know why that is yet.


samuelthoren:
Looks really good, nice work!

Here's a very short review, we can continue reviewing this in January.
- The structure can maybe be improved. Now we have `gui/controlpanel` and `gui/src`, which might be confusing. Maybe we should have a `controlpanel` directory and a `configpanel` directory, and a directory for the shared code or something like that.
- Make it possible to fetch the parameters that are loaded to ATOS, right now it shows the default parameters from the repo.
- Some input about the tabs:
  -  Is it possible to set the default directory that opens when you press `Browse...` to `~/.astazero/ATOS/` instead?
  - When loading for instance an `.xosc`-file, does this it send the full path or only the name? We append `~/.astazero/ATOS/osc/` in EsminiAdapter and if it sends the full path this is not going to work. Maybe a feature that should be added in another PR is to being able to load a scenario file that is located anywhere and not only in `~/.astazero/ATOS/osc/`.
  - I think the integration testing tab is a bit unclear. Is it possible to make it more like the pointcloud tab where you can add and subtract items, and maybe select from pre-defined scenarios?

gui/src/public/javascripts/configClient.js
```diff
@@ -0,0 +1,139 @@
...
+  console.log("update form: ", createdForm);
+  // $('#form').jsonForm(createdForm);
+}
+
+// Wait until ACE is loaded
+var itv = window.setInterval(function() {
+  if (window.ace) {
+    var example = 'atos-param-schema';
+    window.clearInterval(itv);
+    const jsonPlaceholder = loadJsonFile(example);
```
victorjarlow:
```suggestion
    const jsonSchema = loadJsonFile(example);
```

gui/src/views/configpanel.pug
```diff
@@ -0,0 +1,22 @@
...
+  .row
+    .col-md-7
+      h2 ATOS Config Panel
+      form#form.well (please wait)
+script(src='jsonform/deps/jquery.min.js')
+script(src='jsonform/deps/underscore.js')
+script(src='jsonform/deps/opt/ace/ace.js')
+script(src='jsonform/deps/opt/ace/mode-json.js')
+script(src='jsonform/lib/jsonform.js')
+script(src='javascripts/configClient.js')
```
victorjarlow:
```suggestion
      form#form.well (please wait)
//TODO find out a way to include these from the client javascript file, either by using a cdn or by sending the files from the webserver.
script(src='jsonform/deps/jquery.min.js')
script(src='jsonform/deps/underscore.js')
script(src='jsonform/deps/opt/ace/ace.js')
script(src='jsonform/deps/opt/ace/mode-json.js')
script(src='jsonform/lib/jsonform.js')
script(src='javascripts/configClient.js')
```

gui/src/public/javascripts/configClient.js
```diff
@@ -0,0 +1,139 @@
...
+            if (serverResponse.success){
+              window.alert('All parameters were successfully set');
+            }
+            else{
+              window.alert('Not all parameters were successfully set');
+              // break;
+            }
+          case "get_config_response":
+            console.log("get config response: ", serverResponse)
+            // generateForm(serverResponse.value);
```
victorjarlow:
```suggestion
```

gui/src/public/javascripts/configClient.js
victorjarlow: also on the ones below

gui/src/public/javascripts/configClient.js
```diff
@@ -0,0 +1,139 @@
...
+      ' JSON Form object.\n\nThe JSON Form library returned:\n' +
+      e + '</pre>');
+    return;
+  }
+};
+
+function updateForm(jsonString){
+  var createdForm = JSON.parse(jsonString);
+  console.log("update form: ", createdForm);
+  // $('#form').jsonForm(createdForm);
```
victorjarlow:
```suggestion
```

gui/src/public/javascripts/configClient.js
```diff
@@ -0,0 +1,139 @@
...
+  $('#form').html('');
+
+  // Render the resulting form, binding to onSubmitValid
+  try {
+    createdForm.onSubmitValid = function (values) {
+      if (console && console.log) {
+        console.log('Values extracted from submitted form', values);
+      }
+      // window.alert('Form submitted. Values object:\n' +
+      //   JSON.stringify(values, null, 2));
```
victorjarlow:
```suggestion
```

gui/src/ros_nodes/config_node.js
```diff
@@ -0,0 +1,126 @@
...
+      "integration_testing_handler": this._integrationTestingHandlerClient
+    };
+    this._dataType_to_parameterType = {
+      "boolean": ParameterType.PARAMETER_BOOL,
+      "number": ParameterType.PARAMETER_DOUBLE,
+      "string": ParameterType.PARAMETER_STRING,
+    }
+    this._getParametersSrvMsg = 'rcl_interfaces/srv/GetParameters';
+    this._journalGetParametersClient = this._config_node.createClient(this._getParametersSrvMsg, '/atos/journal_control/get_parameters');
+    // Start the ros2 event loop
```
victorjarlow, line 41:
```suggestion
//TODO: Currently only journal parameter client implemented..
    this._getParametersSrvMsg = 'rcl_interfaces/srv/GetParameters';
    this._journalGetParametersClient = this._config_node.createClient(this._getParametersSrvMsg, '/atos/journal_control/get_parameters');
    // Start the ros2 event loop
```

gui/src/ros_nodes/config_node.js
```diff
@@ -0,0 +1,126 @@
...
+    this._config_node.spin();
+  };
+
+  getParamService(serviceClient, ros_parameters, node, ws){
+    const wait_duration_s = 1 // Wait duration for service to appear in seconds.
+    if ( !serviceClient.isServiceServerAvailable() || !serviceClient.waitForService(wait_duration_s*1000)) {
```
victorjarlow:
```suggestion
  getParamService(serviceClient, ros_parameters, node, ws){
  //TODO: This is not yet tested. Getting parameter values for nodes from ROS2 does not currently work.
    const wait_duration_s = 1 // Wait duration for service to appear in seconds.
    if ( !serviceClient.isServiceServerAvailable() || !serviceClient.waitForService(wait_duration_s*1000)) {
```

victorjarlow (changes requested):

victorjarlow (approved):
Nice!

PR #624, opened 2024-01-05
Feature/configpanel python

SebastianSvanland:
If you switch to this branch, you need to add a symlink in atos_ws/src to ATOS/gui/ros_gui:
`cd atos_ws/src && ln -s /path/to/ATOS/gui/ros_gui .`
This is because the gui is its own ros package and won't be found by colcon otherwise.

You also need to install NiceGUI:
`pip install nicegui` 

Run the launch_experimental.py to run the gui together with some nodes, the control and config panel should open automatically. Feel free to run more nodes and try it out (make sure to use atos namespace or it won't show)

Some known issues:
- Some nodes don't have their parameters declared as ros parameters yet in this branch but are still displayed because they are in the json schema.
- If you change a parameter and it doesn't successfully set the parameter, it will still display the incorrect parameter (mainly for nodes that satisfy the above description).


gui/ros_gui/gui/configpanel/configpanel.py
```diff
@@ -0,0 +1,357 @@
...
+class ConfigPanelNode(Node):
+    """ This node is responsible for rendering the config panel and setting parameters in ATOS nodes.
+    """
+
+    def __init__(self) -> None:
+        """ Initializes the node, fetches all ros parameters and renders the config panel with these values.
+        """
+        super().__init__('config_panel')
+        schema_path = os.path.join(CONF_PATH, "atos-param-schema.json")
+        self.json_schema = json.load(open(schema_path, "r"))
```
Robert108: Better to use a context manager (with statement) when handling file operations. That will also take care of closing the file (which isn't done here).
We should also do exception handling if the file does not exist

gui/ros_gui/gui/configpanel/configpanel.py
```diff
@@ -0,0 +1,357 @@
...
+                        for node in self.active_node_list:
+                            ui.tab(node.replace("_", " "))
+                with self.splitter.after:
+                    with ui.tab_panels(tabs, value='Home'):
+                        with ui.tab_panel('Home'):
+                            ui.label('Welcome to ATOS config panel!').classes('text-h4')
+                            if self.active_node_list:
+                                ui.label("Select a node to configure. Press the refresh button if you can't find the node you're looking for.")
+                            else:
+                                ui.label("No nodes were discovered. Press the refresh button to try again.")
```
Robert108: For me the server never seem to find the nodes the first time I start the server and I must always press refresh. It would be nice to not need a button like this and that the server can keep track of nodes automatically.

gui/ros_gui/gui/configpanel/configpanel.py
```diff
@@ -0,0 +1,357 @@
...
+                                                ui.input(param_text).bind_value(self.parameters, param_name).on('keydown.enter', lambda result, node=node, param_name=param_name: self.set_parameter(node, param_name, result.sender.value))
+                                            case "double_array":
+                                                ui.input(param_text).bind_value(self.parameters, param_name).on('keydown.enter', lambda result, node=node, param_name=param_name: self.set_parameter(node, param_name, result.sender.value))
+                                            case "string_array":
+                                                ui.input(param_text).bind_value(self.parameters, param_name).on('keydown.enter', lambda result, node=node, param_name=param_name: self.set_parameter(node, param_name, result.sender.value))
+                                            case _:
+                                                ui.label(f"Unsupported type {param_type}").classes('text-red-500')
+                                        ui.tooltip(param_description)
+
+    def init_clients(self, active_node_list) -> dict:
```
Robert108:
```suggestion
    def init_parameter_clients(self, active_node_list) -> dict:
```

gui/ros_gui/gui/configpanel/configpanel.py
```diff
@@ -0,0 +1,357 @@
...
+    def refresh(self, splitter) -> None:
+        """ Refreshes the config panel by fetching all active nodes and their parameters again.
+        """
+        splitter.delete()
+        modules = self.json_schema["modules"].keys()
+        self.active_nodes_and_namespaces = self.get_node_names_and_namespaces()
+        self.active_node_list = [node for node, namespace in self.active_nodes_and_namespaces if node in modules and "atos" in namespace]
+        self.active_node_list = list(set(self.active_node_list)) # Removes duplicates
+
+        self.client_list = self.init_clients(self.active_node_list)
```
Robert108:
```suggestion
        self.client_list = self.init_parameter_clients(self.active_node_list)
```

gui/ros_gui/gui/configpanel/configpanel.py
```diff
@@ -0,0 +1,357 @@
...
+        super().__init__('config_panel')
+        schema_path = os.path.join(CONF_PATH, "atos-param-schema.json")
+        self.json_schema = json.load(open(schema_path, "r"))
+        modules = self.json_schema["modules"].keys()
+        time.sleep(0.5) # Allow time for this node to initialize before discovering other nodes, otherwise it will fail to find the others.
+        self.active_nodes_and_namespaces = self.get_node_names_and_namespaces()
+        self.active_node_list = [node for node, namespace in self.active_nodes_and_namespaces if node in modules and "atos" in namespace]
+        self.active_node_list = list(set(self.active_node_list)) # Removes duplicates
+
+        self.client_list = self.init_clients(self.active_node_list)
```
Robert108:
```suggestion
        self.client_list = self.init_parameter_clients(self.active_node_list)
```

gui/ros_gui/gui/configpanel/configpanel.py
```diff
@@ -0,0 +1,357 @@
...
+                                        ui.tooltip(param_description)
+
+    def init_clients(self, active_node_list) -> dict:
+        """ Initializes the get and set parameter clients for all active nodes.
+
+        Args:
+            active_node_list (list): List of active nodes.
+
+        Returns:
+            client_list (dict): Dictionary of clients for all active nodes.
```
Robert108: returns a variable called "client_list" but docs says its a dict. I would rather introduce python typing and skip adding suffixes like this to variables. That will both make it clear to the reader and limit future typing errors.

gui/ros_gui/gui/configpanel/configpanel.py
```diff
@@ -0,0 +1,357 @@
...
+        """
+        client_list = {}
+        for module in active_node_list:
+            get_params_client = self.create_client(ListParameters, f'/atos/{module}/list_parameters')
+            get_param_value_client = self.create_client(GetParameters, f'/atos/{module}/get_parameters')
+            set_params_client = self.create_client(SetParametersAtomically, f'/atos/{module}/set_parameters_atomically')
+            client_list[module] = {"get_params_client": get_params_client, "get_param_value_client": get_param_value_client, "set_params_client": set_params_client}
+        return client_list
+
+    def get_parameters_list(self, client_list) -> None:
```
Robert108:
```suggestion
    def set_parameters_list(self, client_list) -> None:
```

gui/ros_gui/gui/configpanel/configpanel.py
```diff
@@ -0,0 +1,357 @@
...
+        Args:
+            node_name (str): Name of the node that the parameter belongs to.
+            param_name (str): Name of the parameter.
+        """
+        result = await local_file_picker('~/.astazero/ATOS', multiple=True, show_hidden_files=True)
+        ui.notify(f'You selected {result}')
+        if not result:
+            return
+        self.parameters[param_name] = str(result)
+        self.set_parameter(node_name, param_name, str(result))
```
Robert108:
```suggestion
```

Robert108: duplicate?

gui/ros_gui/gui/configpanel/configpanel.py
```diff
@@ -0,0 +1,357 @@
...
+        Args:
+            future (Future): Future object returned by the set parameters service call (Bool).
+            param_name (str): Name of the parameter.
+        """
+        try:
+            response = future.result()
+        except Exception as e:
+            self.get_logger().info('Service call failed %r' % (e,))
+            return
+        self.get_logger().debug(f'Setting parameter {param_name} was {"successful" if response.result.successful else "unsuccesful"}')
```
Robert108: I think it is enough to let the user be notified on error. The successful case should be presumed.

gui/ros_gui/gui/configpanel/configpanel.py
```diff
@@ -0,0 +1,357 @@
...
+    def set_parameter(self, node_name, param_name, param_value) -> None:
+        """ Sets the value for a specific parameter.
+
+        Args:
+            node_name (str): Name of the node that the parameter belongs to.
+            param_name (str): Name of the parameter.
+            param_value (any): Value of the parameter.
+        """
+        self.get_logger().debug(f'Setting parameter {param_name} in {node_name} to {param_value}')
+        ui.notify(f'Setting parameter {param_name} in {node_name} to {param_value}')
```
Robert108:
```suggestion
```

gui/ros_gui/gui/configpanel/configpanel.py
```diff
@@ -0,0 +1,357 @@
...
+
+        service_timeout_counter = 0
+        client = self.client_list[node_name]["set_params_client"]
+        while not client.wait_for_service(timeout_sec=1.0):
+            service_timeout_counter += 1
+            self.get_logger().debug('service not available, waiting again...')
+            if service_timeout_counter >= MAX_TIMEOUT:
+                ui.notify(f'Service not available after {MAX_TIMEOUT} seconds, please try again later')
+                self.get_logger().info(f'Service not available after {MAX_TIMEOUT} seconds, please try again later')
+                return
```
Robert108: Since we do wait_for_service in multiple places we should extract it to reduce code duplication.

Robert108 (changes requested):

gui/ros_gui/gui/configpanel/configpanel.py
SebastianSvanland: Hmm, strange. I haven't had this issue but I added an automatic refresh if no nodes were found. Let me know if it helps, because I can't test the behavior myself

gui/ros_gui/gui/configpanel/configpanel.py
SebastianSvanland: I renamed the variable, hopefully it's more clear. I was thinking more in general terms when I named it since it's basically a list of clients for anyone outside the developer-world. But I agree it's confusing for a developer (which virtually everyone who's reading the code is) that's thinking about data types.

gui/ros_gui/gui/configpanel/configpanel.py
SebastianSvanland: Can you elaborate on this suggestion. My thought process is that the function is retrieving all parameters from all active nodes, i.e. getting them. It's not setting the parameters at this point.

gui/ros_gui/gui/configpanel/configpanel.py
SebastianSvanland: There is a very minor difference compared to "pick_file" which is that in "local_file_picker()" there is a flag "multiple" indicating if we can select more than one file or not.

gui/ros_gui/gui/configpanel/configpanel.py
SebastianSvanland: I disagree. The other notification was superfluous, but this one is needed imo. From a user perspective, it's nice to get any form of validation that an action has been made (in this case you also get information the action was successful). If the user wouldn't get this response, he/she wouldn't notice the difference between pressing "enter" to submit or just deselecting the input field. For a new user, I also think that it's a bit too much to explicitly write out that you should submit by pressing enter, it's better for them to learn it by performing the action and seeing a response vs no response for other actions (like deselecting)

SebastianSvanland:
I changed the branch that this PR merges into. This branch now includes changes that makes ros2_based_gui outdated so that branch no longer needs to be merge when this PR is approved. This shouldn't mean any major difference in the review though

gui/ros_gui/gui/configpanel/configpanel.py
Robert108: OK,  Then I suggest adding the "multiple" variable to the pick_file function instead.

gui/ros_gui/gui/configpanel/configpanel.py
Robert108: OK, Im persuaded

launch/launch_utils/launch_base.py
```diff
@@ -52,16 +52,16 @@ def get_base_nodes():
...
+        # ExecuteProcess(
+        #     name='control_gui',
+        #     output={'both': 'log'}, #print to log to avoid cluttering the terminal
+        #     cmd=[[
+        #         FindExecutable(name='npm'),
+        #         ' start --prefix ',
+        #         str(control_gui_dir)
+        #     ]],
+        #     shell=True
+        # ),
```
Robert108:
```suggestion

```

gui/ros_gui/gui/configpanel/configpanel.py
Robert108: Hmm maybe **set** was a bad suggestion. The reason for the name change was that a "**get** function" should usually have no side-effects and just return (normally) local data. **fetch** might be a better name indicating that you are going to call services over the network.

Robert108 (changes requested):

gui/ros_gui/gui/configpanel/configpanel.py
SebastianSvanland: Agree, fetch is probably the most accurate. Fixed it in new commit

Robert108 (approved):

gui/gui/configpanel/local_file_picker.py
```diff
@@ -0,0 +1,82 @@
...
+            with ui.row().classes('w-full justify-end'):
+                ui.button('Cancel', on_click=self.close).props('outline')
+                ui.button('Ok', on_click=self._handle_ok)
+        self.update_grid()
+
+    def add_drives_toggle(self):
+        if platform.system() == 'Windows':
+            import win32api
+            drives = win32api.GetLogicalDriveStrings().split('\000')[:-1]
+            self.drives_toggle = ui.toggle(drives, value=drives[0], on_change=self.update_drive)
```
victorjarlow: Can some os agnostic lib be used to handle files, like os or pathlib?

gui/gui/configpanel/local_file_picker.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def update_drive(self):
+        self.path = Path(self.drives_toggle.value).expanduser()
+        self.update_grid()
```
victorjarlow: Doxygen comments are missing for most functions in this file (and other files), making it hard to get an overview of what the functions do.

gui/gui/controlpanel/controlpanel.py
```diff
@@ -0,0 +1,84 @@
...
+                ui.button('Arm', on_click=lambda: self.armPub.publish(Empty()), color='yellow')
+                ui.button('Disarm', on_click=lambda: self.disarmPub.publish(Empty()), color='grey')
+                ui.button('Start', on_click=lambda: self.startPub.publish(Empty()), color='green')
+                ui.button('All Clear', on_click=lambda: self.allClearPub.publish(Empty()), color='blue')
+            with ui.row():
+                ui.label().bind_text_from(self.OBC_state, 'state', backward=lambda n: f'Object Control State: {n}')
+            with ui.row():
+                ui.button('Reset Test Objects', on_click=lambda: self.resetTestObjectsPub.publish(Empty()), color='grey')
+                ui.button('Reload Object Settings', on_click=lambda: self.reloadObjectSettingsPub.publish(Empty()), color='grey')
+
```
victorjarlow: Is it possible to bring in bootstrap to make this look like the old gui?

gui/gui/main.py
```diff
@@ -0,0 +1,58 @@
...
+import sys
+
+import rclpy
+from rclpy.executors import ExternalShutdownException, MultiThreadedExecutor
+from .controlpanel.controlpanel import ControlPanelNode
+from .configpanel.configpanel import ConfigPanelNode
+
+from nicegui import app, ui, ui_run
+
+USE_SSL = sys.argv[1] == "True"
```
victorjarlow: What happens if sys.argv[1] does not exist?

gui/gui/main.py
```diff
@@ -0,0 +1,58 @@
...
+
+def main() -> None:
+    # NOTE: This function is defined as the ROS entry point in setup.py, but it's empty to enable NiceGUI auto-reloading
+    pass
+
+
+def ros_main() -> None:
+    ui.link('Control Panel', '/control')
+    ui.link('Config Panel', '/config')
```
victorjarlow: maybe use nicegui.ui? to make it easy to understand what ui is referred to

gui/gui/main.py
```diff
@@ -0,0 +1,58 @@
...
+    try:
+        executor.spin()
+    except ExternalShutdownException:
+        pass
+    finally:
+        executor.shutdown()
+        control_panel.destroy_node()
+        config_panel.destroy_node()
+
+#Starting the ros node in a thread managed by nicegui. It will restarted with "on_startup" after a reload.
```
victorjarlow:
```suggestion
#Starting the ros node in a thread managed by nicegui. It will be restarted with "on_startup" after a reload.
```

launch/launch_experimental.py
```diff
@@ -28,13 +28,13 @@ def get_experimental_nodes():
...
-            name='osi_adapter',
-            parameters=[files["params"]]
-        ),
+        # Node(
+        #     package='atos',
+        #     namespace='atos',
+        #     executable='osi_adapter',
+        #     name='osi_adapter',
+        #     parameters=[files["params"]]
+        # ),
```
victorjarlow: Should be commented?

package.xml
```diff
@@ -24,6 +24,7 @@
   <depend>nav_msgs</depend>
   <depend>std_srvs</depend>
   <depend>atos_interfaces</depend>
+  <depend>gui</depend>
```
victorjarlow: can it be called maybe atos_gui or sth? gui is a very generic package name

scripts/installation/install_atos.sh
```diff
@@ -13,30 +13,11 @@ ATOS_REPO_PATH="$1"
 source "${ATOS_REPO_PATH}/scripts/installation/install_functions.sh"
 check_command_failed $? "Failed to source ${ATOS_REPO_PATH}/scripts/installation/install_functions.sh"
 
-################################################
-###### Install Control Panel dependencies ######
-################################################
+#############################################################
+###### Install Python based Control Panel dependencies ######
```
victorjarlow: Why is python based relevant?
Is it only control panel? Maybe its more like ATOS GUI Dependencies

victorjarlow (commented):
I think we should add license text to all files

gui/gui/configpanel/local_file_picker.py
SebastianSvanland: I've barely even looked at this file to be honest. I took it straight from [NiceGUI examples ](https://github.com/zauberzeug/nicegui/blob/main/examples/local_file_picker/local_file_picker.py)(not sure why this isn't added in the main library). I just tried if it worked as intended and didn't bother more with it. I also have no idea what's compatible with Windows and not.

gui/gui/configpanel/local_file_picker.py
SebastianSvanland: See above comment. Maybe I should add some text add the top of the file where the code comes from

launch/launch_experimental.py
SebastianSvanland: I'll leave it to @Robert108. It was commented when I branched out from his initial branch. Maybe something clashed?

gui/gui/controlpanel/controlpanel.py
SebastianSvanland: @Robert108

gui/gui/main.py
SebastianSvanland: I tried not giving any argument in the launch-file and it doesn't seem to result in any issues.

package.xml
SebastianSvanland: Agree, I renamed it.

gui/gui/controlpanel/controlpanel.py
Robert108: We could , But should we? I think both the old and the new design/layout could be up for redesign.

victorjarlow: I think this one looks worse than the old one :D

victorjarlow: So my proposition is: yes, if it's easily possible, we can enable bootstrap (to show its possible to integrate this gui with common frontend libs, and improved layout). Also yes, we can think about how to redesign the layout to promote useability (function is prio 1), such that there is no clutter/distractions to take focus away from what is important, safety-wise.

Robert108: I added some Tailwind and Quasar props to show how you could change the appearance. Do you think this looks better?

image
image

PR #628, opened 2024-01-18
Added github actions for build and clang-tidy/format.

SebastianSvanland:
Added some workflows for adding clang-tidy and clang-format, as well as one for build (clang-tidy depends on the build step). Right now, after some discussion with Robert, the clang tools does not block merging since we just started using them, but you can find the output if you go to the workflow's summary and at the bottom you will find two artifacts (zip-files) called clang-format/tidy-result with suggested fixes. Something to note is that ATOS now uses clang compiler instead of gcc.

I also restructured a bit by putting all files related to ATOS' ros package in one folder and keep the rest (config, scripts, plugins etc). This allows for simpler installation without the need for symbolic links to atos_interfaces, atos_gui and potentially more packages in the future. It also simplified the use of clang-tidy and clang-format.


Robert108 (changes requested):
Nice work! But I miss documentation on which linting/formatting rules that are enforced and how I as a developer should go about formatting my code before pushing so that i don't get caught in this check. Creating a "Style guide" or similar in the contributing section would help a lot. (A great feature is to create a git hook that runs before commits, like this https://github.com/pocc/pre-commit-hooks)

Robert108 (approved):

PR #631, opened 2024-02-14
Feature/mqtt bridge refactor

Robert108:
This PR rewrites mqttbridge to use PahoMqttCpp. It also adds subscription possibilities via mqttsubscribercallback


samuelthoren:
194 files changed? It feels like there are more changes than just the MQTT things

SebastianSvanland:
> 194 files changed? It feels like there are more changes than just the MQTT things

Looks like he branched out from one of my branches. Should have been resolved once it was merged (it was merged after your comment) but still doesn't seem like it...

@Robert108 You might need to merge with dev and resolve any potential conflict before we can properly review this PR

atos/package.xml
SebastianSvanland, line 1: I think you might need to update this file: https://github.com/RI-SE/ATOS/blob/a46b2a98c9980bc3af0cf932af87b33562facb8d/atos/package.xml#L28

atos/modules/MQTTBridge/src/mqttbridge.cpp
```diff
@@ -84,22 +65,20 @@ void MqttBridge::setupConnection()
...
 	{
 		RCLCPP_DEBUG(this->get_logger(), "Successfully initialized MQTT connection to broker");
-		this->mqttClient = mqttClient;
 	}
 }
 
 void MqttBridge::onV2xMsg(const V2X::message_type::SharedPtr v2x_msg)
 {
-	if (MQTTClient_isConnected(this->mqttClient))
+	if (mqttClient->isConnected())
```
SebastianSvanland: Is it needed to do this check? I agree that it's unnecessary to create the payload message etc if it's not connected. But right now there are two checks for this: here and in mqttClient->publishMessage() which feels redundant. Maybe choose one.

atos/modules/MQTTBridge/src/mqttbridge.cpp
```diff
@@ -16,27 +16,26 @@ MqttBridge::MqttBridge() : Module(MqttBridge::moduleName),
...
 	declare_parameter("username","");
 	declare_parameter("password","");
 	declare_parameter("topic","");
-	declare_parameter("quality_of_service","1");
+	declare_parameter("quality_of_service", 1);
 
 	get_parameter("broker_ip", brokerIP);
-	get_parameter("pub_client_id", pubClientId);
+	get_parameter("port", port);	
+	get_parameter("client_id", clientId);
```
SebastianSvanland: clientId is not used except in a print and then the mqttclient has an id that isn't overridden

atos/modules/MQTTBridge/src/mqttclient.cpp
```diff
@@ -0,0 +1,83 @@
...
+		.finalize();
+}
+
+
+MQTTClient::~MQTTClient() {
+	if(mqttClient->is_connected()){
+        mqtt::disconnect_options disconnectOpts;
+        disconnectOpts.set_timeout(std::chrono::seconds(30));
+        mqttClient->disconnect(disconnectOpts)->wait();
+    }
```
SebastianSvanland: Tabs are a bit off (at least in this view on github)

atos/modules/MQTTBridge/inc/mqttsubscribercallback.hpp
```diff
@@ -0,0 +1,24 @@
...
+	public:
+		MQTTSubscriberCallback(mqtt::async_client& client, mqtt::connect_options& connectionOptions, std::vector<std::string>& topics);
+		~MQTTSubscriberCallback();
+
+		std::string mqttMessage = "";
+
+	private:
+		mqtt::async_client& client;
+		mqtt::connect_options& connectionOptions;
+		std::vector<std::string> topics;
```
SebastianSvanland, line 18: Why is topics also not a pointer? Am I missing something?

atos/modules/MQTTBridge/src/mqttsubscribercallback.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+MQTTSubscriberCallback::~MQTTSubscriberCallback() {}
+
+
+/**
+ * @brief Callback for when the client has connected, subscribe to the topics.
+ * 
+ * @param cause Cause.
+ */
+void MQTTSubscriberCallback::connected(const std::string& cause) {
```
SebastianSvanland, line 18: I think either something like "onConnectionEstablished" or simply "subscribe" would be a bit more clear. Although "subscribe" is used enough times already imo.

atos/modules/MQTTBridge/src/mqttsubscribercallback.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * @brief Callback for when a message has arrived, set the message member variable so other
+ * classes can access it.
+ * 
+ * @param msg The message received.
+ */
+void MQTTSubscriberCallback::message_arrived(mqtt::const_message_ptr msg) {
+	mqttMessage = msg->to_string();
+}
```
SebastianSvanland, line 33: What happens if there are multiple topics? Will this member variable change to match the latest message in any of the topics?

atos/modules/MQTTBridge/src/mqttbridge.cpp
Robert108: Aggreed, I will remove this one

atos/modules/MQTTBridge/src/mqttbridge.cpp
Robert108: Sharp eyes! Will remove this one as well

atos/modules/MQTTBridge/inc/mqttsubscribercallback.hpp
Robert108: Subscriber callbacks are created per topics that are subscribed. So i don't think i makes much sense to have it as a pointer.

atos/modules/MQTTBridge/src/mqttsubscribercallback.cpp
Robert108: Connected is an overriden function from mqtt::callback

atos/modules/MQTTBridge/src/mqttsubscribercallback.cpp
Robert108: Yes, a little weird I agree.

atos/modules/MQTTBridge/src/mqttclient.cpp
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * @brief Publish a message to a topic.
+ * 
+ * @param topic The topic to publish to.
+ * @param message The message to publish.
+ */
+void MQTTClient::publishMessage(const std::string& topic, const std::string& message) {
+	if (mqttClient->is_connected()) {
+		mqtt::message_ptr pubmsg = mqtt::make_message(topic, message);
+		pubmsg->set_qos(1);
```
victorjarlow: why is qos hardcoded here in particular?

atos/modules/MQTTBridge/src/mqttclient.cpp
```diff
@@ -0,0 +1,83 @@
...
+		.password(mqttPass)
+		.automatic_reconnect(std::chrono::seconds(MIN_RECONNECT_INTERVAL_SECONDS), std::chrono::seconds(MAX_RECONNECT_INTERVAL_SECONDS))
+		.clean_session(REMEMBER_RECONNECT_STATE)
+		.keep_alive_interval(std::chrono::seconds(KEEP_ALIVE_INTERVAL_SECONDS))
+		.finalize();
+}
+
+
+MQTTClient::~MQTTClient() {
+	if(mqttClient->is_connected()){
```
victorjarlow: In other files in this PR this->xxx is used, but here mqttClient is referred to without this prefix.

atos/modules/MQTTBridge/src/mqttclient.cpp
```diff
@@ -0,0 +1,83 @@
...
+	mqttHost(host),
+	mqttPort(port),
+	mqttUser(username),
+	mqttPass(password) {
+
+	// Each client needs a unique ID, adding a timestamp to the end of the ID should be enough
+	auto timeNanoseconds = std::chrono::duration_cast<std::chrono::nanoseconds>(std::chrono::system_clock::now().time_since_epoch()).count();
+	mqttClientID += "_" + std::to_string(timeNanoseconds);
+
+	mqttClient = std::make_unique<mqtt::async_client>(mqttHost + ":" + std::to_string(mqttPort), mqttClientID);
```
victorjarlow: This is the most important data member of this class and it shares the same name as the class. 
Alternative class name: MQTTClientWrapper ? or something to this effect

atos/modules/MQTTBridge/src/mqttclient.cpp
```diff
@@ -0,0 +1,83 @@
...
+	mqttClient->connect(connectionOptions);
+}
+
+
+/**
+ * @brief Setup the MQTT subscription with the topics provided and connect to the broker.
+ * 
+ */
+void MQTTClient::setupSubscription(std::vector<std::string>& mqttTopics) {
+	mqttSubscriberCallback = std::make_unique<MQTTSubscriberCallback>(*mqttClient, connectionOptions, mqttTopics);
```
victorjarlow: Does this mean that a client can only subscribe to 1 topic at a time?

victorjarlow (approved):

atos/modules/MQTTBridge/src/mqttclient.cpp
Robert108: Oversight I would say.

atos/modules/MQTTBridge/src/mqttclient.cpp
Robert108: Very good suggestion! I changed the wrapper class to MQTTClientWrapper

atos/modules/MQTTBridge/src/mqttclient.cpp
Robert108: The callback class creates a subscription for each topic in the vector when the client connects. I renamed the class to setupSubscription**s** to reflect this.

atos/modules/MQTTBridge/src/mqttclient.cpp
Robert108: This could be due to style differences between me and Samuel how wrote a mqtt implementation on which this is based on. Do we have any preference?

atos/modules/MQTTBridge/inc/mqttclientwrapper.hpp
```diff
@@ -0,0 +1,44 @@
...
+
+private:
+  std::unique_ptr<mqtt::async_client> mqttClient;
+  std::unique_ptr<MQTTSubscriberCallback> mqttSubscriberCallback;
+
+  mqtt::connect_options connectionOptions;
+
+  std::string mqttHost;
+  int mqttPort;
+  std::string mqttClientID = "atos";
```
samuelthoren, line 35: If we are going to create more than one object of this class, `mqttClientID`has to be unique otherwise it will crash. It might be a good idea to make this string unique now so that we don't get this problem in the future

atos/modules/MQTTBridge/src/mqttbridge.cpp
```diff
@@ -4,123 +4,94 @@
...
-		catch (std::runtime_error&)
-		{
-			RCLCPP_ERROR(this->get_logger(), "Failed to publish MQTT message");
-		}
-	}
-	else
-	{
-		RCLCPP_ERROR(this->get_logger(), "Received v2x msg while the client is disconnected, dropping msg...");
-	}
+void MqttBridge::onV2xMsg(const V2X::message_type::SharedPtr v2x_msg) {
```
samuelthoren, line 68:
```suggestion
void MqttBridge::onV2xMsg(const V2X::message_type::SharedPtr v2xMsg) {
```
Doesn't follow the convention ;)

atos/modules/MQTTBridge/inc/mqttclientwrapper.hpp
samuelthoren: I saw that this was fixed, ignore previous comment

samuelthoren (approved):

atos/modules/MQTTBridge/src/mqttbridge.cpp
SebastianSvanland: I agree and disagree. This doesn't follow normal cpp convention. However, this follows ROS convention, which is "biased" towards python convention to make use of ROS across languages more unified. There are even times when using ROS that CamelCase causes issues and errors. Thus I think we should stick to ROS convention whenever we use ROS.

PR #634, opened 2024-02-15
Fix paths, Add set -e

Robert108:


SebastianSvanland:
@Robert108 I think this is solved now

Robert108 (commented):
Looks good!

SebastianSvanland (approved):
LGTM

PR #635, opened 2024-03-01
Fix obc state enum

Robert108:


SebastianSvanland:
Hmm, shouldn't we start adding these things to the ROS messages? That way we ensure that e.g. OBC states enum will match in both backend and frontend and for all other python modules.

SebastianSvanland:
@Robert108 Do you think it's a good idea? We should fix this bug asap

Robert108:
@SebastianSvanland Yes that is a good idea! But it could be worth waiting with introducing enum definitions to when we rewrite OBC. I have a feeling we will do a overhaul of these messages then anyway.

SebastianSvanland (approved):
Sounds fair

PR #636, opened 2024-03-01
Generic mqtt2ros pub sub

Robert108:
This PR introduces generic std_msgs::msg::Empty ros publishers in which the user is able to configure, in the params.yaml file, a mapping from a mqtt topic to a ros topic. 

It removes the old wrapper and subscribercallback classes and instead incorporates the mqtt client directly as a member in the bridge. 

This PR has been tried working with the old way of publishing MQTT msgs out from ATOS with predefined ros topics. (DENMs and state changes)


atos/modules/MQTTBridge/inc/Imqtt2ros.hpp
```diff
@@ -0,0 +1,20 @@
+#include <rclcpp/rclcpp.hpp>
+#include <std_msgs/msg/empty.hpp>
+
+/**
+ * @brief Struct containing variables related to a MQTT2ROS connection.
+ */
+struct Mqtt2RosInterface {
+  struct {
+    int qos = 0; ///< MQTT QoS value
```
SebastianSvanland, line 11: Shouldn't there be a topic here?

conf/conf/params.yaml
```diff
@@ -17,12 +17,33 @@ atos:
...
+      ros2mqtt:           # Object specifying which ROS topics to map to which MQTT topics
+        ros_topics:
+          - placeholder2
+        placeholder2:
+          mqtt_topic: placeholder2
+          advanced:
+            mqtt:
+              qos: 1
+            ros:
+              queue_size: 1
```
SebastianSvanland, line 46: Nest the ros topic with mqtt topic directly under "ros_topics"

atos/modules/MQTTBridge/src/mqttbridge.cpp
```diff
@@ -15,21 +15,134 @@ MqttBridge::MqttBridge()
...
+        mqtt2ros.mqtt.qos = qos_param.as_int();
+      }
+
+      // mqtt2ros[k]/advanced/ros/queue_size
+      rclcpp::Parameter queue_size_param;
+      if (get_parameter(
+              fmt::format("mqtt2ros.{}.advanced.ros.queue_size", mqtt_topic),
+              queue_size_param)) {
+        mqtt2ros.ros.queue_size = queue_size_param.as_int();
+      }
```
SebastianSvanland, line 98: Is it possible to read the parameter tree "mqtt2ros.{mqtt_topic}" once and then read each parameter from there afterwards instead? Would make it cleaner imo

SebastianSvanland (commented):
Rather complicated but I guess there's not much to do about it. I left a few comments

atos/modules/MQTTBridge/inc/Imqtt2ros.hpp
Robert108: I'm a bit hestiant to add the topic name here. I'll add a comment here instead how this struct how this struct is intended to be used.

atos/modules/MQTTBridge/src/mqttbridge.cpp
Robert108: I looked in to this and couldn't find a good way to declare the parameters without a string list.

conf/conf/params.yaml
Robert108: This did not work due the same reason as the problem of declaring the parameters inside the mqttbridge.cpp file :(

SebastianSvanland (approved):

PR #638, opened 2024-03-14
Fix/update GitHub actions

Robert108:


SebastianSvanland (approved):
Looks good, ATOS now builds locally again as well

PR #641, opened 2024-04-09
Update network settings in compose file

Robert108:


ceciliahernqvist:
This change gives me Failed TCP connection after Connect:
image

Same scenario/procedure works fine with ATOS dev branch.
I am not running Docker Desktop.

Robert108:
@ceciliahernqvist Yepp, that is expect. When running with bridge mode you will need to specify you docker interface address (often called docker0 with an ip of 172.17.0.1 on the host) in the VehicleCatalog instead of the loop back address 127.0.0.1 which you can do when running in host mode.

ceciliahernqvist:
Aa okay, makes sense then.

ceciliahernqvist (approved):

Robert108:
@ceciliahernqvist I changed this some more and updated the documentation on using docker network. Now you don't need to know the IPs of the isoObject containers or the host IP, you just need to specify hostnames and they will be resolved automatically.

PR #653, opened 2024-09-18
Update test scenario for failing test

Robert108:


SebastianSvanland (approved):
In a future PR we should probably update the scenario with new descriptive event names, since we are no longer restricted to the prefixed event names.

PR #655, opened 2024-10-11
Start on more xosc actions

Robert108:


SebastianSvanland:
I would approve this, but the CI tests doesn't seem to go through

SebastianSvanland (approved):

PR #661, opened 2025-02-11
Migrate to Boost SML state machine

mnil:


samuelthoren:
Nice work! I have not reviewed the code yet, but I tested it and it seems to work in the same way as before

victorjarlow:
Also tested with CBLA scenario with two object scenarios and 1000 hertz MONR rate, and it worked like before, got one segfault in isoobject though.

victorjarlow (approved):

PR #677, opened 2026-04-20
Apply formatting

samuelthoren:
Apply formatting by running `python3 format.py --all-files`. Nothing else has been changed in this PR except running that command. It's hard to see what has actually changed in other PR's when it also applies the formatting in those.

The formatting now fails because some python scripts, which I don't think we should change for this PR because of the risk of changing some behaviour


JonathanAhlstedt (approved):
I looked quickly through. Looks good

PR #679, opened 2026-04-21
Update esmini version

samuelthoren:
The version for esmini that was used is no longer on the release page: https://github.com/esmini/esmini/releases?page=3.

Changed to https://github.com/esmini/esmini/releases/tag/v2.37.17 instead.

Also updates versions in `requirements.txt`


JonathanAhlstedt (approved):
"Thumbs up"

PR #680, opened 2026-04-21
Fix/integration test

samuelthoren:
Add fix for that the iso-object needs to start before ATOS, also add support for manually starting the integration test


JonathanAhlstedt (approved):
Looks great

PR #683, opened 2026-05-12
Add Fleet Manager build integration

sepast:


atos/CMakeLists.txt
sepast, line 1: Added a `WITH_TRUCK_OBJECT_CONTROL` CMake option and appended the module to `ENABLED_MODULES` when the flag is enabled.
This was needed so the new TruckObjectControl module can be built selectively without changing the default build path for existing environments.

atos/common/roschannels/monitorchannel.hpp
sepast, line 1: Replaced `tf2::toMsg` / `tf2::fromMsg` helper usage with direct quaternion field assignment/construction and added the missing `tf2::Matrix3x3` include.
This was done to make the quaternion conversions compile reliably across ROS distros where the `tf2_geometry_msgs` helper API availability differs.

atos/common/trajectory.cpp
sepast, line 1: Added explicit `tf2` linear math headers and replaced message-conversion helpers with direct quaternion component copies.
Updated `Path` header stamping so the path header and pose timestamps share the same time base.
These changes were needed to restore compilation under newer ROS environments and keep emitted path timestamps internally consistent.

atos/common/util.c
sepast, line 1: Added missing return statements, corrected debug `fprintf` calls, fixed a bad string terminator assignment, and corrected mismatched debug format arguments.
These changes were needed to resolve hard compiler errors and non-void fallthrough warnings that blocked the build on stricter toolchains.

atos/common/trajectory.hpp
sepast, line 1: Added a direct `#include <rclcpp/rclcpp.hpp>`.
This was needed because the implementation now depends on `rclcpp::Time` / `rclcpp::Duration` being declared from the public header path, rather than relying on indirect includes.

atos/modules/IntegrationTesting/src/integrationtesting.cpp
sepast, line 1: Added a null/failed-service guard before reading the object-control-state response.
This was needed to prevent the integration test process from crashing when the service call fails during startup or scenario-loading delays.

atos/modules/EsminiAdapter/src/esminiadapter.cpp
sepast, line 1: Made `SE_ReportObjectPos` conditional on the ROS build environment: Humble keeps the timestamp argument, other environments use the shorter signature.
This was needed because the installed esmini headers differ between environments, so one fixed call signature could not compile everywhere.

atos/modules/ObjectControl/src/objectcontrol.cpp
sepast, line 1: Increased the waits for trajectory, IP, and origin service responses, and extended the overall scenario-loading timeout.
This was needed because CI/container startup is slower than local runs, and the previous timeouts were causing false scenario-load failures during integration testing.

scripts/installation/dependencies.txt
sepast, line 1: Added `libssl-dev`.
This was needed to satisfy SSL/TLS-related build or runtime dependencies introduced by the current branch's environment and bridge usage.

scripts/integration_testing/run_scenario_test.py
sepast, line 1: Disabled `foxglove_bridge` for the integration test launch and increased the test timeout from 30s to 60s.
This was needed because `foxglove_bridge` was failing in the test container while not being relevant to the scenario test itself, and the previous timeout was too short for slower CI runs.

atos/modules/EsminiAdapter/src/esminiadapter.cpp
```diff
@@ -221,8 +221,12 @@ void EsminiAdapter::reportObjectPosition(const Monitor::message_type::SharedPtr
 	auto speed = monr->velocity.twist.linear;
 
 	// Reporting to Esmini
-	int timestamp = 0; // Not really used according to esmini documentation
+#if ROS_HUMBLE
+	float timestamp = 0.0f; // Not really used according to esmini documentation
 	SE_ReportObjectPos(esminiObjectId, timestamp, pos.x, pos.y, pos.z, yaw, pitch, roll);
+#else
+	SE_ReportObjectPos(esminiObjectId, pos.x, pos.y, pos.z, yaw, pitch, roll);
+#endif
```
samuelthoren: Is this really depending on the ROS version? It should be the version of `esmini` that we use, I think? I think it was correct how it was before the change here

atos/modules/ObjectControl/src/objectcontrol.cpp
samuelthoren: Is this really needed? The CI has passed on the latest merge to `dev`

atos/modules/TruckObjectControl/inc/atostrucksimulator.hpp
```diff
@@ -0,0 +1,59 @@
...
+	int start_index_					 = 0;
+	double initial_speed_kmh_			 = 0.0;
+	double target_speed_kmh_			 = 40.0;
+	double acceleration_mps2_			 = 2.0;
+	double publish_hz_					 = 5.0;
+	bool loop_path_						 = true;
+	bool ignore_warning_speed_commands_	 = false;
+
+	double current_distance_m_ = 0.0;
+	double current_speed_mps_  = 0.0;
```
samuelthoren: Add `m_` prefix to all member variables and remove `_`

atos/modules/TruckObjectControl/inc/atostrucksimulator.hpp
```diff
@@ -0,0 +1,59 @@
...
+
+class AtosTruckSimulator : public rclcpp::Node {
+public:
+	AtosTruckSimulator();
+
+private:
+	struct GeoPoint {
+		double lat		  = 0.0;
+		double lon		  = 0.0;
+		double distance_m = 0.0;
```
samuelthoren: Distance to what? Specify

atos/modules/TruckObjectControl/inc/atostrucksimulator.hpp
samuelthoren: Also should we be able to change these through some config file? Right now they are hardcoded, and it takes time to rebuild everything if we want to test with other values

atos/modules/TruckObjectControl/inc/truckobjectcontrol.hpp
```diff
@@ -0,0 +1,145 @@
...
+#include <set>
+#include <string>
+#include <thread>
+#include <unordered_map>
+#include <vector>
+
+class TruckObjectControl : public rclcpp::Node {
+public:
+	TruckObjectControl();
+	~TruckObjectControl() override;
```
samuelthoren, line 26: Implement or `delete` the rest of the functions as well https://en.cppreference.com/cpp/language/rule_of_three

atos/modules/TruckObjectControl/inc/truckobjectcontrol.hpp
```diff
@@ -0,0 +1,145 @@
...
+	double cot_timeout_seconds_			 = 2.0;
+	int cot_tcp_port_					 = 8114;
+	std::string cot_tcp_bind_address_	 = "0.0.0.0";
+	bool cot_tls_require_client_cert_	 = false;
+	std::string cot_tls_cert_path_		 = "";
+	std::string cot_tls_key_path_		 = "";
+	std::string cot_tls_ca_path_		 = "";
+	bool cot_tls_enabled_				 = false;
+	std::string trajectory_geojson_path_ = "";
+	std::string default_path_name_		 = "";
```
samuelthoren: Also fix the member variable names for this class

atos/modules/TruckObjectControl/src/atostrucksimulator.cpp
```diff
@@ -0,0 +1,455 @@
...
+		return configured_path;
+	}
+
+	std::vector<fs::path> candidates;
+	candidates.emplace_back(fs::current_path() / "conf" / "conf" / kGeoJsonName);
+	candidates.emplace_back(fs::current_path() / ".." / "conf" / "conf" / kGeoJsonName);
+
+	if (const char* home = std::getenv("HOME")) {
+		candidates.emplace_back(fs::path(home) / "atos_ws" / "src" / "atos" / "conf" / "conf" / kGeoJsonName);
+		candidates.emplace_back(fs::path(home) / "Documents" / "repos" / "ATOS" / "conf" / "conf" / kGeoJsonName);
```
samuelthoren: This path feels hardcoded to your computer

atos/modules/TruckObjectControl/src/truckobjectcontrol.cpp
```diff
@@ -0,0 +1,1496 @@
...
+#include <set>
+#include <sstream>
+#include <stdexcept>
+#include <vector>
+
+using std::placeholders::_1;
+using json = nlohmann::json;
+
+namespace {
+constexpr size_t kReceiveBufferSize		   = 4096;
```
samuelthoren:
```suggestion
constexpr std::size_t kReceiveBufferSize		   = 4096;
```

atos/modules/TruckObjectControl/src/truckobjectcontrol.cpp
```diff
@@ -0,0 +1,1496 @@
...
+constexpr int kAcceptPollSleepMs		   = 100;
+constexpr int kTlsHandshakeRetrySleepMs	   = 10;
+constexpr int kTlsReadPollTimeoutMs		   = 1000;
+constexpr int kTlsIoRetrySleepMs		   = 10;
+constexpr int kTlsWriteMaxTransientRetries = 100;
+constexpr const char* kDefaultGeoJsonName  = "RuralRoad_center_of_driving_lane_ccw.geojson";
```
samuelthoren: `std::string_view` can be used here instead https://en.cppreference.com/cpp/string/basic_string_view

atos/modules/TruckObjectControl/src/truckobjectcontrol.cpp
```diff
@@ -0,0 +1,1496 @@
...
+
+double degToRad(double value) {
```
samuelthoren:
```suggestion
double degToRad(const double value) {
```

Check where `const` can be applied for all functions

atos/modules/TruckObjectControl/src/truckobjectcontrol.cpp
```diff
@@ -0,0 +1,1496 @@
...
+		return configured_path;
+	}
+
+	std::vector<fs::path> candidates;
+	candidates.emplace_back(fs::current_path() / "conf" / "conf" / kDefaultGeoJsonName);
+	candidates.emplace_back(fs::current_path() / ".." / "conf" / "conf" / kDefaultGeoJsonName);
+
+	if (const char* home = std::getenv("HOME")) {
+		candidates.emplace_back(fs::path(home) / "atos_ws" / "src" / "atos" / "conf" / "conf" / kDefaultGeoJsonName);
+		candidates.emplace_back(fs::path(home) / "Documents" / "repos" / "ATOS" / "conf" / "conf" /
```
samuelthoren: Also here (hardcoded path?)

atos/modules/EsminiAdapter/src/esminiadapter.cpp
sepast: You’re right. The conditional is tied to the wrong dependency.

atos/modules/TruckObjectControl/inc/atostrucksimulator.hpp
sepast: the field now clearly means cumulative distance along the loaded
  trajectory path.

atos/modules/TruckObjectControl/src/atostrucksimulator.cpp
sepast: deleted ~/Documents/repos/ATOS/...

atos/modules/ObjectControl/src/objectcontrol.cpp
sepast: Yes, the tests failed if the timeouts not where increased.

atos/modules/EsminiAdapter/src/esminiadapter.cpp
```diff
@@ -221,8 +221,12 @@ void EsminiAdapter::reportObjectPosition(const Monitor::message_type::SharedPtr
 	auto speed = monr->velocity.twist.linear;
 
 	// Reporting to Esmini
-	int timestamp = 0; // Not really used according to esmini documentation
+#ifdef ESMINI_REPORT_OBJECT_POS_HAS_TIMESTAMP
+	float timestamp = 0.0f; // Not really used according to esmini documentation
 	SE_ReportObjectPos(esminiObjectId, timestamp, pos.x, pos.y, pos.z, yaw, pitch, roll);
+#else
+	SE_ReportObjectPos(esminiObjectId, pos.x, pos.y, pos.z, yaw, pitch, roll);
+#endif
```
samuelthoren: I don't think we should change anything here. In `install_deps.sh` we have `esmini v2.37.17` which expects `SE_ReportObjectPos(esminiObjectId, timestamp, pos.x, pos.y, pos.z, yaw, pitch, roll);` with the timestamp: https://github.com/esmini/esmini/blob/3c64169fc2d2b2bf2fcbd426017ae856ffb6de57/EnvironmentSimulator/Libraries/esminiLib/esminiLib.cpp#L1147

atos/modules/EsminiAdapter/CMakeLists.txt
```diff
@@ -44,6 +56,10 @@ target_include_directories(${ESMINI_ADAPTER_TARGET} PUBLIC SYSTEM
 	${esminiRMLib_INCLUDE_DIRS}
 )
 
+if(ESMINI_REPORT_OBJECT_POS_HAS_TIMESTAMP)
+	target_compile_definitions(${ESMINI_ADAPTER_TARGET} PRIVATE ESMINI_REPORT_OBJECT_POS_HAS_TIMESTAMP)
+endif()
+
```
samuelthoren: I think we should revert this as well, same reason as the other comment

atos/modules/TruckObjectControl/inc/atostrucksimulator.hpp
```diff
@@ -0,0 +1,59 @@
...
+	rclcpp::Subscription<std_msgs::msg::String>::SharedPtr m_speed_command_sub;
+
+	rclcpp::Time m_last_step_time;
+
+	bool loadTrajectoryPath();
+	bool ensureTcpConnected();
+	void closeTcp();
+	void pollTcpCommands();
+	void simulationStep();
+	bool pointAtDistance(double distance_m, double& lat, double& lon, double& course_deg, int& path_index) const;
```
samuelthoren, line 55:
```suggestion
	bool pointAtDistance(double distance_m, double lat, double lon, double course_deg, int path_index) const;
```

samuelthoren: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f16-for-in-parameters-pass-cheaply-copied-types-by-value-and-others-by-reference-to-const

sepast: Don't understand the comment, 

bool pointAtDistance(double distance_m, double lat, double lon, double course_deg, int path_index) const;

would pass lat, lon, course_deg, and path_index by value, so any changes inside pointAtDistance(...)  would be lost when the function returns

scripts/installation/install_deps.sh
sepast, line 1: The change was needed because the script was hardcoded to ROS_DISTRO=humble

atos/modules/TruckObjectControl/inc/atostrucksimulator.hpp
samuelthoren: Ah I see now that you modify the parameters in the function, then we need to keep it. Let's keep it as it is

samuelthoren (approved):

PR #685, opened 2026-06-18
Fix ATOS install and Docker runtime environment

sepast:


Dockerfile
sepast, line 1: switched the image build to the supported setup_atos.sh flow so Docker uses the same install contract as native setup

setup_atos.sh
sepast, line 1: made setup follow the active supported ROS distro and coordinate dependency install plus ATOS install consistently

scripts/installation/install_deps.sh
sepast, line 1: - fixed dependency installation for Ubuntu 24.04/PEP 668
      - creates the ATOS Python venv
      - installs Python requirements into that venv
      - keeps ROS Python tooling available during builds

scripts/installation/dependencies.txt
sepast, line 1: added python3-openssl so ROS launch code using OpenSSL works in the container/system Python environment

scripts/run_atosfleetmanagement.sh
sepast, line 1: activates the ATOS Python venv before launching ATOS so runtime Python dependencies are available

docker-compose-test.yml
sepast, line 1: activates the ATOS venv before running integration tests so launch-time Python dependencies are present

docker-compose.yml
sepast, line 1: activates the ATOS venv before starting the standard Docker runtime

docker-compose-bridge.yml
sepast, line 1: activates the ATOS venv before starting the bridge-enabled Docker runtime

PR #689, opened 2026-07-03
Check distance to path

samuelthoren:
If trucks are more than 7.5 meters outside of the path, then it won't affect the other trucks

### DIFF
--- a/atos/modules/TruckObjectControl/inc/truckobjectcontrol.hpp
+++ b/atos/modules/TruckObjectControl/inc/truckobjectcontrol.hpp
@@ -38,6 +38,7 @@ private:
 
 	struct TruckState {
 		double distance_along_trajectory_m = 0.0;
+		double distance_to_path_m		   = 0.0;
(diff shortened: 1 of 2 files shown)


sepast (approved):
Tested with simulators:
   source /opt/ros/jazzy/setup.bash
  colcon build --packages-select atos --symlink-install --cmake-args DWITH_TRUCK_OBJECT_CONTROL=ON
  source install/setup.bash

  Normal simulation test:

  ros2 launch atos launch_atosfleetmanagement.py \
    foxbridge:=False \
    insecure:=True \
    with_truck_simulator:=True

  Off-path simulator test:

  ros2 launch atos launch_atosfleetmanagement.py \
    foxbridge:=False \
    insecure:=True \
    with_truck_simulator:=True \
    simulator_2_lateral_offset_m:=10.0

when offset is set to 10 m is the truck ignored and is not affecting the trucks with no offset.

PR #694, opened 2026-07-08
Permission for script and build packages

samuelthoren:
Add correct permissions for the script and build the `TruckObjectControl` package


sepast (approved):
Where are the !isAhead changes?

